PR #1, opened 2017-03-08
Update focus-ring.js

robdodson:
just fixing the repo name in the comment


yoavweiss:

PR #4, opened 2017-03-08
Clean up polyfill code; add discussion of default modality to explainer.

alice:
@robdodson PTAL

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,29 @@ To deal with this:
 Additionally, :focusring matches non-native elements as if they were
 native button elements.
 
+### Note: Styling non-native elements which should always match `focus-ring`
+
+This is not currently part of the spec,
+but a mechanism is needed to explain the ability of native text fields
+to match `:focus-ring` regardless of how focus arrived on the element.
+
+Two different strategies may enable this
+("`-bikeshed`" placeholder indicates that these names are by no means final!):
+
+1. Add a new keyword value to the outline shorthand that represents whatever the default UA `::focus-ring` is. Then authors can do:
+
+        :focus {
+            outline: platform-default-focus-outline-style-bikeshed;
+        }
+
+2. Add a new CSS property that controls "keyboard modality" vs non-"keyboard modality" behavior, e.g.
+
+        my-custom-texty-element {
+            should-show-focus-on-mouse-click-bikeshed: on-bikeshed;
+        }
+
+It may make sense to implement both of these strategies.
+
 ## Example heuristic
 
 The heuristic used to decide the current modality should not be defined
--- a/README.md
+++ b/README.md
@@ -83,16 +83,16 @@ show the default focus ring on click for this element:
 1. Add a new keyword value to the outline shorthand that represents whatever the default UA `::focus-ring` is. Then authors can do:
 
         custom-texty-element:focus {
-            outline: platform-default-focus-outline-style-bikeshed;
+            outline: platform-default-focus-outline-style-foo;
         }
 
 2. Add a new CSS property that controls "keyboard modality" vs non-"keyboard modality" behavior, e.g.
 
         custom-texty-element {
-            show-focus-ring-bikeshed: always | keyboard-only;
+            show-focus-ring-foo: always | keyboard-only;
         }
 
-_("`-bikeshed`" placeholder indicates that these names are by no means final!)_
+_("`-foo`" placeholder indicates that these names are by no means final!)_
 
 While either of these primitives would suffice,
 having both would provide more flexibility for authors.
--- a/README.md
+++ b/README.md
@@ -59,26 +59,43 @@ native button elements.
 
 ### Note: Styling non-native elements which should always match `focus-ring`
 
+
 This is not currently part of the spec,
 but a mechanism is needed to explain the ability of native text fields
 to match `:focus-ring` regardless of how focus arrived on the element.
 
-Two different strategies may enable this
-("`-bikeshed`" placeholder indicates that these names are by no means final!):
+Consider an author creating a custom element, `custom-texty-element`,
+which they believe should show a focus ring on mouse click.
+By default, the default `:focus-ring` user agent style
+will not show a focus ring when this element receives focus via mouse click.
+However, if the author were to style the element via `:focus`,
+they could not recreate the browser's default `outline` style reliably:
+
+```css
+custom-texty-element:focus {
+   outline: ???;
+}
+```
+
+Either of the following two new primitives would allow the author to
+show the default focus ring on click for this element:
 
 1. Add a new keyword value to the outline shorthand that represents whatever the default UA `::focus-ring` is. Then authors can do:
 
-        :focus {
+        custom-texty-element:focus {
             outline: platform-default-focus-outline-style-bikeshed;
         }
 
 2. Add a new CSS property that controls "keyboard modality" vs non-"keyboard modality" behavior, e.g.
 
-        my-custom-texty-element {
-            should-show-focus-on-mouse-click-bikeshed: on-bikeshed;
+        custom-texty-element {
+            show-focus-ring-bikeshed: always | keyboard-only;
         }
 
-It may make sense to implement both of these strategies.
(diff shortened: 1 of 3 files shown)


marcoscaceres:
@alice, I might be wrong, but it seems you have not yet joined the WICG as a Google rep. Please visit the following link, and click "join this group". That should guide you and give you the right permissions.  https://www.w3.org/community/wicg/ 

In the mean time, I'll mark this as non-substitantive.

marcoscaceres:

README.md
```diff
@@ -57,6 +57,29 @@ To deal with this:
...
+This is not currently part of the spec,
+but a mechanism is needed to explain the ability of native text fields
+to match `:focus-ring` regardless of how focus arrived on the element.
+
+Two different strategies may enable this
+("`-bikeshed`" placeholder indicates that these names are by no means final!):
+
+1. Add a new keyword value to the outline shorthand that represents whatever the default UA `::focus-ring` is. Then authors can do:
+
+        :focus {
```
robdodson: suggest using `:focus-ring`

README.md
```diff
@@ -57,6 +57,29 @@ To deal with this:
...
+            outline: platform-default-focus-outline-style-bikeshed;
+        }
+
+2. Add a new CSS property that controls "keyboard modality" vs non-"keyboard modality" behavior, e.g.
+
+        my-custom-texty-element {
+            should-show-focus-on-mouse-click-bikeshed: on-bikeshed;
```
robdodson: I feel like if we do this then why have `:focus-ring` at all? Not saying we _shouldn't_ do this, just that it would probably supersede whatever `:focus-ring` has to offer

demo/index.html
```diff
@@ -32,7 +32,9 @@
     <h1>Keyboard-only focus test</h1>
     <p>
       This page contains various html controls and demonstrates a proposal for only drawing a focus ring when the keyboard is being used.
-      Navigate through the interactive elements various ways (using mouse, switch to keyboard, try it on a touch device). You should see a green focus indicator (or comment out green focus ring style to see default focus ring) on elements below differently based on whether you interact with them via keyboard or mouse.  If you are using the mouse and get lost, at any time you can tab/shift-tab to find your focus indicator.
+      Navigate through the interactive elements various ways (using mouse, switch to keyboard, try it on a touch device).
```
robdodson: "Navigate through the interactive elements **in** various ways"

src/focus-ring.js
```diff
@@ -1,67 +1,112 @@
-/* https://github.com/WICG/focusring */
+/* https://github.com/WICG/focus-ring */
```
robdodson: oops I merged my own PR that had this same change
https://github.com/WICG/focus-ring/pull/1

src/focus-ring.js
```diff
@@ -1,67 +1,112 @@
...
 document.addEventListener('DOMContentLoaded', function() {
-    var hadKeyboardEvent = false,
-        keyboardModalityWhitelist = [ 'input:not([type])',
+    var hadKeyboardEvent = false;
```
robdodson: Suggest using `let` and `const` throughout so we can be cool and hip. 🏂

src/focus-ring.js
```diff
@@ -1,67 +1,112 @@
...
+     * On `keydown`, set `hadKeyboardEvent`, to be removed 100ms later if there
+     * are no further keyboard events.  The 100ms throttle handles cases where
+     * focus is redirected programmatically after a keyboard event, such as
+     * opening a menu or dialog.
+     */
+    function onKeyDown() {
         hadKeyboardEvent = true;
+        // `activeElement` defaults to document.body if nothing focused,
+        // so check the active element is actually focused.
         if (document.activeElement.matches(':focus')) {
```
robdodson: should this be `matcher.call(document.activeElement, ':focus')

src/focus-ring.js
```diff
@@ -1,67 +1,112 @@
...
-            el.setAttribute('data-focus-ring-added', '');
-        },
-        removeFocusRingClass = function(el) {
-            if (!el.hasAttribute('data-focus-ring-added'))
-                return;
-            el.classList.remove('focus-ring');
-            el.removeAttribute('data-focus-ring-added')
-        };
+                                      '[role=textbox]' ].join(',');
+    var matcher = (function() {
```
robdodson: would you mind renaming this to `matches`? And adding a comment linking to the MDN post https://developer.mozilla.org/en-US/docs/Web/API/Element/matches

when I think of "matcher" my brain automatically assumes that this is already bound to a specific element. I have gone back and reread this chunk of code like 5 times and I still can't shake that feeling 😋

src/focus-ring.js
```diff
@@ -1,67 +1,112 @@
...
+        keyboardThrottleTimeoutID = setTimeout(function() {
             hadKeyboardEvent = false;
+            keyboardThrottleTimeoutID = 0;
         }, 100);
-    }, true);
+    }
 
-    document.body.addEventListener('focus', function(e) {
+    /**
+     * On `focus`, add the `focus-ring` class to the target if necessary.
```
robdodson: replace `if necessary` with `if the element matches a selector in the keyboardModalityWhitelist whitelist`

src/focus-ring.js
```diff
@@ -1,67 +1,112 @@
-/* https://github.com/WICG/focusring */
+/* https://github.com/WICG/focus-ring */
 document.addEventListener('DOMContentLoaded', function() {
-    var hadKeyboardEvent = false,
-        keyboardModalityWhitelist = [ 'input:not([type])',
+    var hadKeyboardEvent = false;
+    var keyboardThrottleTimeoutID = 0;
+
+    var keyboardModalityWhitelist = [ 'input:not([type])',
```
robdodson: This could use a comment explaining why the whitelist exists. That it's for elements we want to _always_ display a `:focus-ring` for

src/focus-ring.js
alice: doesn't that mean it work work in some browsers and we'll have to go through all that pain again?

src/focus-ring.js
alice: I had that briefly but `matches` by itself looks too much like it should be a _list_ of matches (results) to me. `matcher` conveys that it's a function, IMO.

src/focus-ring.js
alice: derp

src/focus-ring.js
alice: Not accurate:
- Will also add it if focus came from keyboard and the element doesn't match the whitelist
- Will not add it if it was already there.

robdodson: could you add all of the above to the comment? 😸

alice: Why? They're implementation details.

robdodson: I guess I see the potential scenarios as critical pieces of information.

```
On `focus`, add the `focus-ring` class if one of the following scenarios is met:
- If focus came from the keyboard.
- If focus occurred before the keyboardThrottle expired.
- If the `target` is in the keyboardModalityWhitelist.
```

README.md
robdodson: oh sorry I thought this was more about our whole `outline: ua` discussion. Like if someone says `:focus { outline: none }` then there's no way for `:focus-ring { }` to ever get the default UA styles back

README.md
robdodson: OOOOOOOOH. got it. yes that would be nice. I think the way the use case was pitched didn't click in my head... also it's 1am

You may be able to rephrase it as:

```
In addition to `:focus-ring`, there are additional, speculative, missing primitives which may make working with focus easier.

1. An element should have a way to restore the UA default outline. Because each UA has a different default style, there's no easy way to restore the outline if a prior selector has overridden it (short of UA sniffing).

2. An element should have a way to express when it should display a focus ring. On keyboard focus, on mouse focus, or on both.

These could help explain the behavior of native text fields...
```

src/focus-ring.js
alice: Ok you convinced me.

README.md
alice: Reworded.

PR #10, opened 2017-03-25
Add older IE support

joe-watkins:
Hi :)

I've tweaked the Polyfill to reach back a bit further with IE.

1. Added `.gitignore` file
2. Defined HTML class and attribute as var: `focus-ring`, `data-focus-ring-added`
2. Went with `indexOf()` instead of `.contains('focus-ring')`
2. Created `_addClass()` & `_removeClass()` helper functions to replace `el.classList.add()` & `el.classList.remove()`
3. Killed the `.matches(':focus')` method on line 58 -- no old IE support and felt redundant.
4. Cleaned up HTML in demo.. something in there was causing old IE to freak out. Added doctype, and IE meta.

### DIFF
--- a/src/focus-ring.js
+++ b/src/focus-ring.js
@@ -1,6 +1,8 @@
 /* https://github.com/WICG/focus-ring */
 document.addEventListener('DOMContentLoaded', function() {
     var hadKeyboardEvent = false,
+        focusRingClass = 'focus-ring',
+        focusRingAttr = 'data-focus-ring-added',
         keyboardModalityWhitelist = [ 'input:not([type])',
                                       'input[type=text]',
                                       'input[type=number]',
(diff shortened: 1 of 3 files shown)


src/focus-ring.js
```diff
@@ -29,22 +31,31 @@ document.addEventListener('DOMContentLoaded', function() {
 	    }
 	    return triggers;
 	},
+        _addClass = function(el, htmlClass) {
+            el.className += ' '+htmlClass;   
+        },
+        _removeClass = function(el, htmlClass) {
```
robdodson: Do you know the support level of RegExp in older IE? It might be nicer to do something like:

```
this.element.className = this.element.className.replace(regExp(name), '');
```

based on the polyfill example in https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#Polyfill

src/focus-ring.js
```diff
@@ -29,22 +31,31 @@ document.addEventListener('DOMContentLoaded', function() {
 	    }
 	    return triggers;
 	},
+        _addClass = function(el, htmlClass) {
```
robdodson: because we're declaring these inside of a function I don't think you need to underscore their names

src/focus-ring.js
```diff
@@ -29,22 +31,31 @@ document.addEventListener('DOMContentLoaded', function() {
...
-            el.classList.remove('focus-ring');
-            el.removeAttribute('data-focus-ring-added')
+            _removeClass(el,focusRingClass);
+            el.removeAttribute(focusRingAttr)
         };
 
     document.body.addEventListener('keydown', function() {
         hadKeyboardEvent = true;
-        if (document.activeElement.matches(':focus')) {
+        if (document.activeElement) {
```
robdodson: This change seems ok to me. cc @alice for a 2nd opinion

src/focus-ring.js
joe-watkins: Sounds good - I've removed those underscores :)

src/focus-ring.js
joe-watkins: Good call! I've snagged the regex helper function from the classList Polyfill and followed their lead. Feels cleaner :) (And because I'll never understand Regex :) ha!)

yoavweiss:
Hey @joe-watkins, could you join the WICG in order to appease the IPR bots? Thanks!

yoavweiss:
@joe-watkins potentially helpful link: https://www.w3.org/community/wicg/participants

joe-watkins:
@robdodson This branch is way off from from gh-pages, different worlds,.. and merge conflicts are pretty gnarly. I've signed up over at WICG (tnx @yoavweiss !) which may help w/future PRs. I'm going to pull latest `gh-pages` and inject the IE changes from that and create a new PR.

Issue #6, opened 2017-03-24
Progressive Enhancement recommendation


joe-watkins:
Hello :) Great work on this!

I'm wondering if you'd be interested in including another example in the demo/docs that demonstrated a progressive enhancement approach where we 'detect' JS and if it is 'present' we kill outlines.

If JS fails or there is an error, no elements will have focus rings.

I've [created an example of this](https://github.com/joe-watkins/focus-ring/commit/b434cd4cae39dfd9b965589da7d337727350ee4c) for review.

joe-watkins:
.. or maybe even consider having JavaScript add a class `focus-ring-polyfill` that you can descend from:

`.focus-ring-polyfill :focus:not(.focus-ring) { outline: none; }`

With something like that JavaScript could fail or have an error and default focus states would be intact. See #8

robdodson:
Adding a class to the body works, but I'm concerned about causing really large style recalcs on older browsers.

> Note: It used to be the case that if you changed a class on -- say -- the body element, that all the children in the page would need to have their computed styles recalculated. Thankfully that is no longer the case; some browsers instead maintain a small collection of rules unique to each element that, if changed, cause the element’s styles to be recalculated. That means that an element may or may not need to be recalculated depending on where it is in the tree, and what specifically got changed.
https://developers.google.com/web/fundamentals/performance/rendering/reduce-the-scope-and-complexity-of-style-calculations#reduce_the_number_of_elements_being_styled

I did a test in Chrome and things are fine, we only recalc 1 element if it's a mouse click, or 2 elements if we're tabbing and the focus ring is moving. But I'm less optimistic about older versions of IE. @paullewis were you referring to specific browsers in that comment? I'd love to know the versions if so.

robdodson:
Another option is the polyfill could inject the styles itself. That way it only does it if JS is working. I'm not in love with the idea of something injecting CSS because I prefer having total control over that, but it might be a decent compromise in this case. Or we could even make it configurable.

robdodson:
We're going with the class on body approach as discussed in https://github.com/WICG/focus-ring/issues/8

I'm going to close this issue so we can track the work in #8.

PR #11, opened 2017-03-25
Add HTML class via JavaScript - outline:none CSS descends

joe-watkins:
Hello ;)

The concern raised in #6 is that, in supporting browsers, all focusable elements will not have a focus ring should JavaScript fail or is not present with this CSS:

``` CSS
:focus:not(.focus-ring) {
  outline: none;
}
```

In the name of progressive enhancement, I'd like to propose that the Polyfill adds an HTML class to the `html` element and then CSS descends from that.

With this approach we are fairly certain JavaScript is present, not disabled, and/or there are not any errors.

**Here is the process:**

1. HTML class of `` is added to the `<html>` tag.

``` JavaScript
document.documentElement.className += " focus-ring-polyfill";
```

2. The CSS descends from that class:

``` CSS
.focus-ring-polyfill :focus:not(.focus-ring) {
   outline: none;
}
```


joe-watkins:
Going to kill this while we figure out #6 options -- will create a new PR if we come up with something for it. :)

PR #17, opened 2017-03-31
Fix clearTimeout to use correct timeout id

patrickhlauke:


robdodson:
good catch, thank you 😄

PR #18, opened 2017-04-02
Add old IE support

joe-watkins:
Here's a PR to support for older versions of IE.

- Adding support for older versions of IE that don't support `classList`
- Borrowed regex helper from [classlist Polyfil](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#Polyfill)
- Cleaned up demo HTML

### DIFF
--- a/src/focus-ring.js
+++ b/src/focus-ring.js
@@ -5,6 +5,7 @@ document.addEventListener('DOMContentLoaded', function() {
         keyboardThrottleTimeoutID = 0,
         focusRingClass = 'focus-ring',
         focusRingAttr = 'data-focus-ring-added',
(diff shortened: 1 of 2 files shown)


joe-watkins:

src/focus-ring.js
```diff
@@ -59,10 +72,14 @@ document.addEventListener('DOMContentLoaded', function() {
      * @param {Element} el
      */
     function addFocusRingClass(el) {
-        if (el.classList.contains('focus-ring'))
+        if (el.className.indexOf(focusRingClass) != -1)
```
joe-watkins: I'm thinking regex here would be better to match exact.

robdodson:
Fixed by #27

PR #23, opened 2017-04-06
Fix Element.matches()'s selectorString syntax

jwyung:
This fixes a typo in not closing the `:not()` pseudo-class in the
selectorString.


alice:
Thanks for catching that!

PR #24, opened 2017-04-06
Add support for browsers without support for classLast and matches

kloots:
This PR adds support for browsers lacking support for `classLast` and `matches` (e.g. Safari 9). We need this for Slack as Safari 9 is our baseline. This PR also adds linting and a build process.


robdodson:
Hey thank you for the PR! I just landed from a flight where I was working on something similar. I think I'd like to add rollup to the build process and export a UMD version of the library and a rolled up ES module (ES5 code + ES6 imports). I'll open a separate PR tomorrow to show you what I've got and we can discuss.

kloots:
Sounds good to me Rob. Let me know if you end up not having time to work on that and I can take over getting the PR up.

robdodson:
Fixed by #27

PR #30, opened 2017-05-31
Add a focusring custom event

kloots:
This PR adds a `focusring` custom event, triggered on the element to which the `focus-ring` class is to be applied. This allows projects further control over when a focus outline is rendered for a given element. 

For example, at Slack we're using the focus-ring library but prefer to _only_ see a focus outline when the user is navigating via the keyboard using either `Tab` or `Shift` + `Tab`. With this change we'd be able to add a single, document-level listener to manage this:

```
document.addEventListener('focusring', function(e) {
  if (e.detail.srcEvent.keyCode != 9) {
    e.preventDefault();
  }
}, true);
```


alice:
Hm, I don't think this belongs in the polyfill - this behaviour is not specced, and I don't think it could be specced for a pseudo-selector.

kloots:
@alice i'm not adding a behavior, just adding an event. What troubles you about adding an event?

kloots:
@alice the only alternatives I could see for addressing this bug were:
1. Changing the polyfill such that only `Tab` and `Shift` + `Tab` keypress result in focus-ring being applied—since by default in the browser those are the only keys that result in focus anyway.
2. Allowing for the polyfill to be initialized with a config so people can supply the keys that should result in the `focus-ring` class being applied.

kloots:
All things considered, providing an event feels the most with-the-grain of how existing DOM works.

alice:
Firing the event is the behaviour I was referring to. What troubles me is that this is intended as a polyfill for a pseudo-selector, and pseudo-selectors don't fire events when they get matched.

I don't think this is a bug per se - it's a case of desiring different behaviour from the pseudo-selector. The specifics of *when* `:focus-ring` should match are specifically left vague, so that browsers may choose what they believe is best for their users (so in any case, using native `:focus-ring` would never give you this level of control). Currently, this polyfill matches the behaviour of a native `<button>` element.

I filed https://github.com/WICG/focus-ring/issues/33 to continue this discussion - could you comment there about why you prefer this matching behaviour?

kloots:
Closing this PR as I've addressed the need for this solution in [PR #34](https://github.com/WICG/focus-ring/pull/34) after careful consideration with @alice.

Issue #33, opened 2017-06-02
When to match :focus-ring:?


alice:
Currently we model the behaviour of `<button>`, which (in Chrome) shows a focus ring when it receives focus after a keyboard event, or when a key is pressed while it is focused.

Another option would be to show a focus ring *only* when focus is *received* from a keyboard event. This would be a break from existing behaviour, but may be input into browsers' eventual decision of what behaviour to model.

kloots:
@alice as I am currently working on putting this polyfill to work in the Slack codebase my feedback is: not _every_ focus event following a keydown event should result in a focus ring being drawn. As such, the behavior as currently implemented results in the focus ring showing up unexpectedly/or at times when it is not desired. 

An example from Slack: a user presses a keyboard shortcut (`Cmd` + `.`) to show or hide the right-hand sidebar. We move focus to the sidebar when it is shown, so users have the _option_ to navigate it with the keyboard. However, that focus ring is only useful should the user actually begin navigating via the keyboard. And not everyone opens the sidebar with the intention or need to navigate its content via the keyboard. As such, I end up having to add more CSS to suppress the focus ring style for this instance/state.

The library would be much more useful, reliable and easier to work with if I could simply rely on the focus style provided by the focus-ring polyfill to only appear when a control receives focus as a result of the user navigating to it via the default methods provided by the browser: `Tab` & `Shift` + `Tab`. That way I could simply have:

```
// Suppress focus by default
:focus {
  outline: none;
}

// Show focus for text inputs
input[type=text],
textarea {
  outline: 3px solid blue;
}

// Show focus when navigating via the keyboard
.focus-ring {
  outline: 3px solid blue;
}
```
This is much more with the grain of default browser behavior and solves the real problem I have: users and designers only want to see focus when either navigating via Tab OR if you are typing in a text box.

alice:
As a counter-argument, if both keyboard shortcuts and pointer-based options are available to access a piece of UI, I think using the keyboard shortcut _would_ be a good signal that the user is intending to continue using the keyboard to interact with that UI. In fact, predominantly keyboard users are almost certainly more likely to use keyboard shortcuts which result in moving focus, aren't they?

kloots:
I can tell you with certainty from using this library with Slack that assumption is not correct. :) In fact, I wouldn't have raised this issue if it was correct. :) At this point I'm dealing with way more "why am I seeing this focus ring" bug reports than "I was expecting to see a focus ring and didn't" reports.

[...]

PR #34, opened 2017-06-05
Fix issue number #33

kloots:
This PR addresses [issue #33](https://github.com/WICG/focus-ring/issues/33) by refining the heuristics for when the focus-ring class is applied: if the user navigates to it via `Tab` or `Shift` + `Tab`.

In addition, this PR also:

1. Removes use of `data-focus-ring-added` attribute
2. Removes the need for use of `dom-matches` polyfill dependency 
3. Removes the need for `setTimeout()` when handling the `keydown` event.

### DIFF
--- a/dist/focus-ring.js
+++ b/dist/focus-ring.js
@@ -170,80 +170,25 @@ ClassList.prototype.toggle = function (token, force) {
   return (typeof force == 'boolean' ? force : !hasToken);
 };
 
-/**
- * Determine if a DOM element matches a CSS selector
- *
- * @param {Element} elem
- * @param {String} selector
- * @return {Boolean}
- * @api public
- */
-
-function matches(elem, selector) {
-  // Vendor-specific implementations of `Element.prototype.matches()`.
-  var proto = window.Element.prototype;
-  var nativeMatches = proto.matches ||
-      proto.mozMatchesSelector ||
-      proto.msMatchesSelector ||
-      proto.oMatchesSelector ||
-      proto.webkitMatchesSelector;
-
-  if (!elem || elem.nodeType !== 1) {
-    return false;
-  }
-
-  var parentElem = elem.parentNode;
-
-  // use native 'matches'
-  if (nativeMatches) {
-    return nativeMatches.call(elem, selector);
-  }
-
-  // native support for `matches` is missing and a fallback is required
-  var nodes = parentElem.querySelectorAll(selector);
-  var len = nodes.length;
-
-  for (var i = 0; i < len; i++) {
-    if (nodes[i] === elem) {
-      return true;
-    }
-  }
-
-  return false;
-}
-
-/**
- * Expose `matches`
- */
-
-var index$1 = matches;
-
 /* https://github.com/WICG/focus-ring */
 document.addEventListener('DOMContentLoaded', function() {
   var hadKeyboardEvent = false;
-  var keyboardThrottleTimeoutID = 0;
   var elWithFocusRing;
-
-  // These elements should always have a focus ring drawn, because they are
-  // associated with switching to a keyboard modality.
-  var keyboardModalityWhitelist = [
-    'input:not([type])',
-    'input[type=text]',
-    'input[type=search]',
-    'input[type=url]',
-    'input[type=tel]',
-    'input[type=email]',
-    'input[type=password]',
-    'input[type=number]',
-    'input[type=date]',
-    'input[type=month]',
-    'input[type=week]',
-    'input[type=time]',
-    'input[type=datetime]',
-    'input[type=datetime-local]',
-    'textarea',
-    '[role=textbox]',
-  ].join(',');
+  var inputTypesWhitelist = {
+    'text': true,
+    'search': true,
+    'url': true,
+    'tel': true,
+    'email': true,
+    'password': true,
+    'number': true,
+    'date': true,
+    'month': true,
+    'week': true,
+    'time': true,
+    'datetime': true,
+    'datetime-local': true,
+  };
 
   /**
    * Computes whether the given element should automatically trigger the
@@ -253,7 +198,22 @@ document.addEventListener('DOMContentLoaded', function() {
    * @return {boolean}
    */
   function focusTriggersKeyboardModality(el) {
-    return index$1(el, keyboardModalityWhitelist) && index$1(el, ':not([readonly])');
+    var type = el.type;
+    var tagName = el.tagName.toLowerCase();
+
+    if (tagName == 'input' && inputTypesWhitelist[type] && !el.readonly) {
+      return true;
+    }
+
+    if (tagName == 'textarea' && !el.readonly) {
+      return true;
+    }
+
+    if (el.getAttribute('role').toLowerCase() == 'textbox') {
+      return true;
+    }
+
+    return false;
   }
 
   /**
@@ -265,7 +225,6 @@ document.addEventListener('DOMContentLoaded', function() {
     if (index(el).contains('focus-ring'))
       return;
     index(el).add('focus-ring');
-    el.setAttribute('data-focus-ring-added', '');
     // Keep a reference to the element to which the focus-ring class is applied
     // so the focus-ring class can be restored to it if the window regains
     // focus after being blurred.
@@ -278,10 +237,7 @@ document.addEventListener('DOMContentLoaded', function() {
    * @param {Element} el
    */
   function removeFocusRingClass(el) {
-    if (!el.hasAttribute('data-focus-ring-added'))
-      return;
     index(el).remove('focus-ring');
-    el.removeAttribute('data-focus-ring-added');
   }
 
   /**
@@ -289,31 +245,41 @@ document.addEventListener('DOMContentLoaded', function() {
    * are no further keyboard events.  The 100ms throttle handles cases where
    * focus is redirected programmatically after a keyboard event, such as
    * opening a menu or dialog.
+   * @param {Event} e
    */
-  function onKeyDown() {
-    hadKeyboardEvent = true;
+  function onKeyDown(e) {
+    if (e.altKey || e.ctrlKey || e.metaKey)
+      return;
+
+    var isTabKey = e.keyCode == 9;
+    var isTabNavigation = isTabKey || isTabKey && e.shiftKey;
+
+    if (!isTabNavigation)
+      return;
+
     // `activeElement` defaults to document.body if nothing focused,
     // so check the active element is actually focused.
-    if (index$1(document.activeElement, ':focus'))
-      addFocusRingClass(document.activeElement);
-    if (keyboardThrottleTimeoutID !== 0)
-      clearTimeout(keyboardThrottleTimeoutID);
-    keyboardThrottleTimeoutID = setTimeout(function() {
-      hadKeyboardEvent = false;
-      keyboardThrottleTimeoutID = 0;
-    }, 100);
+    var activeElement = document.activeElement;
+    if (activeElement.tagName == 'body') {
+      return;
+    }
+
+    addFocusRingClass(activeElement);
+    hadKeyboardEvent = true;
   }
 
   /**
    * On `focus`, add the `focus-ring` class to the target if:
-   * - a keyboard event happened in the past 100ms, or
-   * - the focus event target triggers "keyboard modality" and should always
-   *   have a focus ring drawn.
+   * - a keydown event preceded the focus event, meaning the target received
+   *   focus as a result of keyboard navigation.
+   * - the event target is an element that will likely require interaction
+   *   via the keyboard (e.g. a text box)
    * @param {Event} e
    */
   function onFocus(e) {
     if (hadKeyboardEvent || focusTriggersKeyboardModality(e.target))
       addFocusRingClass(e.target);
+      hadKeyboardEvent = false;
   }
 
   /**
@@ -325,7 +291,7 @@ document.addEventListener('DOMContentLoaded', function() {
   }
 
   /**
-   * When the window regains focus, restore the focus-ring class to the element 
+   * When the window regains focus, restore the focus-ring class to the element
    * to which it was previously applied.
    */
   function onWindowFocus() {
@@ -334,9 +300,9 @@ document.addEventListener('DOMContentLoaded', function() {
     }
   }
 
-  document.body.addEventListener('keydown', onKeyDown, true);
-  document.body.addEventListener('focus', onFocus, true);
-  document.body.addEventListener('blur', onBlur, true);
+  document.addEventListener('keydown', onKeyDown, true);
+  document.addEventListener('focus', onFocus, true);
+  document.addEventListener('blur', onBlur, true);
   window.addEventListener('focus', onWindowFocus, true);
 });
 
--- a/dist/focus-ring.js
+++ b/dist/focus-ring.js
@@ -261,6 +261,9 @@ document.addEventListener('DOMContentLoaded', function() {
    * @param {Event} e
    */
   function onFocus(e) {
+    if (e.target == document)
+      return;
+
     if (hadKeyboardEvent || focusTriggersKeyboardModality(e.target)) {
       addFocusRingClass(e.target);
       hadKeyboardEvent = false;
@@ -272,6 +275,9 @@ document.addEventListener('DOMContentLoaded', function() {
    * @param {Event} e
    */
   function onBlur(e) {
+    if (e.target == document)
+      return;
+
     removeFocusRingClass(e.target);
   }
 
--- a/dist/focus-ring.js
+++ b/dist/focus-ring.js
@@ -199,19 +199,16 @@ document.addEventListener('DOMContentLoaded', function() {
    */
   function focusTriggersKeyboardModality(el) {
     var type = el.type;
-    var tagName = el.tagName.toLowerCase();
+    var tagName = el.tagName;
 
-    if (tagName == 'input' && inputTypesWhitelist[type] && !el.readonly) {
+    if (tagName == 'INPUT' && inputTypesWhitelist[type] && !el.readonly)
       return true;
-    }
 
-    if (tagName == 'textarea' && !el.readonly) {
+    if (tagName == 'TEXTAREA' && !el.readonly)
       return true;
-    }
 
-    if (el.getAttribute('role').toLowerCase() == 'textbox') {
+    if (el.getAttribute('role').toLowerCase() == 'textbox')
       return true;
-    }
 
     return false;
   }
@@ -225,6 +222,7 @@ document.addEventListener('DOMContentLoaded', function() {
     if (index(el).contains('focus-ring'))
       return;
     index(el).add('focus-ring');
+    el.setAttribute('data-focus-ring-added', '');
     // Keep a reference to the element to which the focus-ring class is applied
     // so the focus-ring class can be restored to it if the window regains
     // focus after being blurred.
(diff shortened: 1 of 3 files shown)


alice:
Mostly looks good, but I'd like to get at least @bkardell @robdodson or @marcysutton to take a look before we merge, since this is a big behaviour change.

src/focus-ring.js
```diff
@@ -61,42 +67,49 @@ document.addEventListener('DOMContentLoaded', function() {
...
+   * - a keydown event preceded the focus event, meaning the target received
+   *   focus as a result of keyboard navigation.
+   * - the event target is an element that will likely require interaction
+   *   via the keyboard (e.g. a text box)
    * @param {Event} e
    */
   function onFocus(e) {
     if (hadKeyboardEvent || focusTriggersKeyboardModality(e.target))
       addFocusRingClass(e.target);
+      hadKeyboardEvent = false;
```
bkardell: always use {} and save yourself the hassle is my motto :)

src/focus-ring.js
```diff
@@ -61,42 +67,49 @@ document.addEventListener('DOMContentLoaded', function() {
...
+    if (!isTabNavigation)
+      return;
+
     // `activeElement` defaults to document.body if nothing focused,
     // so check the active element is actually focused.
-    if (matches(document.activeElement, ':focus'))
-      addFocusRingClass(document.activeElement);
-    if (keyboardThrottleTimeoutID !== 0)
-      clearTimeout(keyboardThrottleTimeoutID);
-    keyboardThrottleTimeoutID = setTimeout(function() {
```
kloots: Isn't adding a focus ring superfluous in that case as the UI change alone is an indicator?

dist/focus-ring.js
```diff
@@ -265,7 +225,6 @@ document.addEventListener('DOMContentLoaded', function() {
     if (index(el).contains('focus-ring'))
       return;
     index(el).add('focus-ring');
-    el.setAttribute('data-focus-ring-added', '');
```
alice: Oh, why is this being removed here?

src/focus-ring.js
kloots: @bkardell agreed. All `{}` all the time. Done. Only didn't previously as previously the code didn't use them and I thought that may have been the preferred style by @alice or @robdodson.

src/focus-ring.js
```diff
@@ -61,42 +67,49 @@ document.addEventListener('DOMContentLoaded', function() {
...
+    addFocusRingClass(activeElement);
+    hadKeyboardEvent = true;
   }
 
   /**
    * On `focus`, add the `focus-ring` class to the target if:
-   * - a keyboard event happened in the past 100ms, or
-   * - the focus event target triggers "keyboard modality" and should always
-   *   have a focus ring drawn.
+   * - a keydown event preceded the focus event, meaning the target received
```
kloots: Done.

src/focus-ring.js
```diff
@@ -61,42 +67,49 @@ document.addEventListener('DOMContentLoaded', function() {
...
+   * @param {Event} e
    */
-  function onKeyDown() {
-    hadKeyboardEvent = true;
+  function onKeyDown(e) {
+    if (e.altKey || e.ctrlKey || e.metaKey)
+      return;
+
+    var isTabKey = e.keyCode == 9;
+    var isTabNavigation = isTabKey || isTabKey && e.shiftKey;
```
kloots: It does indeed. Previous approach was more out of the desire to have the code be explicit & self-documenting about when we match. But if you don't like it, then consider it changed.

src/focus-ring.js
alice: I think it depends what ends up being focused. If it's a widget you might immediately want to interact with, I think it might be disorienting to hide the focus ring. Using enter to activate a button *is* a stronger signal than using a keyboard shortcut.

dist/focus-ring.js
kloots: Because I couldn't work out why it was necessary in the first place. Isn't it superfluous? Doesn't the presence of the `focus-ring` class give us enough?

src/focus-ring.js
kloots: I'm not entirely clear on the use case you are advocating for. Would you mind explaining in more detail? Perhaps with a real-life example.

dist/focus-ring.js
alice: It disambiguates an explicit, author-added `focus-ring` class (to allow always matching `.focus-ring`) from a programmatically added one. See line 281 for where it's used. It was added in https://github.com/WICG/focus-ring/commit/8d005806af5dc98b1ee2661ff764aaeda2150018

src/focus-ring.js
alice: I don't have a real life example, unfortunately, but I'm imagining something like a custom date picker which pops up some extra UI when you press a button.

Another thought: this also means `:focus-ring` wouldn't match for UI which supports directional navigation.

src/focus-ring.js
```diff
@@ -52,8 +52,10 @@ document.addEventListener('DOMContentLoaded', function() {
    * @param {Element} el
    */
   function addFocusRingClass(el) {
-    if (classList(el).contains('focus-ring'))
+    if (classList(el).contains('focus-ring')) {
```
alice: Can we please revert these?

kloots: Can do.

dist/focus-ring.js
kloots: What's a use case for an author-added `focus-ring` class? And _why_ would we want to support that if this is a polyfill for `:focus-ring` and there would be no similar pattern once this hopefully is implemented in browsers. Wouldn't we want to discourage authors adding `focus-ring`?

src/focus-ring.js
kloots: ...and done.

dist/focus-ring.js
alice: Check out the CL for context.

src/focus-ring.js
kloots: Isn't directional navigation in the same category as a screen reader's virtual navigation, and therefore the visual indicator is provided by the host service (e.g. like VoiceOver's black outline).

dist/focus-ring.js
alice: Essentially, we need *some* mechanism (as @bkardell alluded to on the bug) for opting things in to the always-match-focus-ring behaviour which elements on the whitelist get. For the purposes of this polyfill, this is what we came up with. For the pseudo-selector, we need to come up with a good alternative still.

src/focus-ring.js
alice: I mean author-implemented directional navigation as opposed to "spatial navigation" as a UA feature using some kind of D-pad. 

The example I just saw was in Google Slides, where you go to add a new slide and you get a (in my case at least) 3x4 grid of slide layout options that you can choose between.

dist/focus-ring.js
kloots: I understand the problem now (the ability to ensure a given element always has a focus ring applied), however what I don't understand _why_ this polyfill needs to account for that when this can be accomplished by authors leveraging the natural cascade and/or specificity rules in CSS.

```
// hide focus by default
:focus {
  outline: none;
}
// focus ring for keyboard navigation
.focus-ring {
  outline: solid 2px blue;
}

// always show focus outline around a specifc control
button.default-action:focus {
  outline: solid 2px blue;
}
```

This way the polyfill is designed to be with-the-grain of the eventual pseudo class implementation.

src/focus-ring.js
bkardell: Not all keyboard users are also using AT, if you had a menu of menu items for example you want them to show the ring, right?

dist/focus-ring.js
bkardell: Can you explain how this is more 'with-the-grain' of the eventual pseudo? I'm not following.

src/focus-ring.js
kloots: @bkardell wouldn't a focus ring be superfluous for menu items? Consider the menubar in Mac OS—you don't get both the selection highlight + a focus ring when navigating via the up & down arrow keys, you simply see the selection state change (blue background).

I don't know how we could account for all author-defined directional navigation while at the same time not matching too aggressively (as we are now). Hence [my comment in issue #33 indicating we should be more conservative with this polyfill](https://github.com/WICG/focus-ring/issues/33#issuecomment-305924855) and allow authors to decide when to show the focus ring when they are implementing author-defined keyboard shortcuts or navigation.

For example, take your menu use case. Any menu I have ever implemented with keyboard accessibility has code that looks like this:

```
function onKeyDown(e) {
  if (e.keyCode == 40) {
    $(e.target).next().focus();
    $(e.target).next().addClass('selected');
  }
}
```
In such cases I'm already explicitly managing both where focus should move and what it should look like via adding a class.

dist/focus-ring.js
alice: In any case, this is orthogonal to the main point of this PR, so maybe let's revert it here and continue the discussion on a bug?

kloots: @bkardell I don't think this polyfill should give authors abilities they won't have with the actual `:focus-ring` psuedo. Therefore, once actual support for `:focus-ring` lands in browsers the only change authors need to make to their CSS is changing their rules from `.focus-ring` to `:focus-ring`. Hence, more with the grain.

alice: Can we file a bug for this discussion and revert this change in the context of this PR?

kloots: @alice [bug is filed](https://github.com/WICG/focus-ring/issues/35). And this PR has been updated to restore `data-focus-ring-added`.

dist/focus-ring.js
```diff
@@ -278,42 +237,49 @@ document.addEventListener('DOMContentLoaded', function() {
    * @param {Element} el
    */
   function removeFocusRingClass(el) {
-    if (!el.hasAttribute('data-focus-ring-added'))
```
marcysutton: I see your commit to add these back in, but for some reason the diff is showing these are still removed.

src/focus-ring.js
```diff
@@ -72,31 +76,37 @@ document.addEventListener('DOMContentLoaded', function() {
    * are no further keyboard events.  The 100ms throttle handles cases where
```
marcysutton: If throttling is going away this comment should probably be updated, too.

dist/focus-ring.js
```diff
@@ -278,42 +237,49 @@ document.addEventListener('DOMContentLoaded', function() {
...
+   * @param {Event} e
    */
-  function onKeyDown() {
-    hadKeyboardEvent = true;
+  function onKeyDown(e) {
+    if (e.altKey || e.ctrlKey || e.metaKey)
+      return;
+
+    var isTabKey = e.keyCode == 9;
+    var isTabNavigation = isTabKey || isTabKey && e.shiftKey;
```
bkardell: this seems a bit of a weird and'ing and or'ing?

marcysutton:
Initially I wasn't sure if suppressing focus on keyboard shortcuts was a good idea, but you can still override the focus style using `outline-width`...so this works. Here's a demo. https://codepen.io/marcysutton/pen/ryqKmd

dist/focus-ring.js
kloots: @bkardell it is right in the source code, I just forgot to re-build the dist. Will do.

dist/focus-ring.js
kloots: @marcysutton Sorry about that! I forgot to re-build the dist. Fixed now.

src/focus-ring.js
kloots: @marcysutton nice eagle eye. Thank you! Fixed.

dist/focus-ring.js
kloots: @bkardell fixed now.

kloots:
@alice wanted to check in on this PR and if you've got any outstanding concerns that need to be addressed. @marcysutton and @bkardell have reviewed and I've addressed their issues. (@marcysutton, @bkardell either of you have any additional comments, or are you good?) I'm still waiting on a review from @robdodson.

alice:
Apologies, I've been in meetings all day. Yeah, would like to get @robdodson's 2c but he's out till next week.

alice:
@marcysutton 
> Initially I wasn't sure if suppressing focus on keyboard shortcuts was a good idea, but you can still override the focus style using outline-width...so this works. 

Can you expand on how this plays into the context of whether focus has come from tab/shift tab or not?

kloots:
@alice I _think_ @marcysutton's comment with respect to "suppressing focus on keyboard shortcuts" was related to [the thread you, @bkardell and I had going yesterday with respect to  author-implemented directional navigation](https://github.com/WICG/focus-ring/pull/34#discussion_r120222234).

marcysutton:
Yes, it was. With this version of the polyfill, tab/shift+tab worked as I expected and the keyboard shortcut case was still achievable with :focus.

alice:
Sure, the part I don't follow is how the `outline` style applies in the keyboard shortcut case - do you mean you would listen for the keyboard shortcut and then apply that style?

marcysutton:
[Here's a demo](https://codepen.io/marcysutton/pen/ryqKmd)–I got a button to show a focus style after manually sending focus there via keyboard shortcut by using a combination of `:focus` and `:focus-ring`. I'm not sure if that's the use case you were talking about, but the change in this PR did suppress the focus style for that navigation pattern before I added a `:focus` override.

alice:
Right - that button also gets a focus ring if you just click it. So the part I'm wondering about is how we handle the case where we *do* want the button to get a focus ring on keyboard shortcuts, without reverting to simply `:focus`-style behaviour where it also gets a focus ring on click.

marcysutton:
Oh yeah, duh! This doesn't actually work then. And it highlights another issue–what if I want to show the focus style for focus management? I use that pattern often to show keyboard users where they are on the screen after sending focus to a new area. That style is suppressed with this change.

kloots:
@alice @marcysutton I have to admit we're so far into this thread I've lost the original context with respect to:

1. The problem as perceived by @marcysutton 
2. If the perceived problem is new/different from one we've already hashed out in previous comments for this PR or in issue #33.

So, my apologies for not following along with you both.

_If_ @marcysutton's concern is how to indicate focus when the user switches contexts (e.g. into a dialog, into a menu, or a side panel), I was under the impression @alice had already addressed that [in her summary for why the recommendation to UAs is `:focus-ring` should only match on `Tab` or `Shift` + `Tab`](https://github.com/WICG/focus-ring/issues/33#issuecomment-306004359):

> "If a keyboard shortcut is likely to move focus, it is probably either showing some new UI, or it is specifically designed to assist keyboard users, in which case a manual focus style is likely to be applied."

@alice's comment is inline with my experience with any webapp I've ever worked on; the code managing a context switch—especially when moving to a dialog or menu—is also explicitly setting focus programmatically to the next control representing the next action the user is likely to take, and also applying some class to draw attention to said control.

In the case of a dialog this is most often either:
1. The button confirming an action (e.g. Highlighting "Ok" in a confirmation dialog)
2. Moving focus into the next most logical text box in a form in a dialog so the user can start typing

In either of these use case, authors can leverage the cascade to ensure the focus ring is visible in response to programatic focus following a context switch:

```
/* Hide focus ring by default */
:focus {
  outline: none
}

/* Show a focus outline when navigating via Tab */
:focus-ring {
  outline: 3px solid blue;
}

/* Ensure the default action button in dialogs has a focus outline */
.default-action-button:focus {
  background: green;
  outline: 3px solid blue;
}
```

In the case of a popup menu I've never had a need to render a focus outline as the selection change made via keyboard navigation is always intended to render to match the selection state on mouseover.

Anyway, sorry for the long message. I _hope_ this resolves this issue (presuming this is the issue you are actually talking about).

marcysutton:
The reason I bring up the focus management use case is that it's impacted by this change. Sadly, I wouldn't use `:focus-ring` if it didn't provide adequate styling hooks for my use cases–I can continue to use What Input for that.

The problem I hoped `:focus-ring` would solve is that custom elements and focus management wrappers with `tabindex` should persist focus styles only after keyboard interaction (including "indirectly", per the spec). I encountered that same gotcha in my demo in this thread when I tried to override the behavior in this PR with a `:focus` style on a custom element. In my opinion, that puts us right back at the core of the problem.

For example, this code persists the focus style on click if the element is a custom control with `tabindex="0"` (now included in my [demo](https://codepen.io/marcysutton/pen/ryqKmd)):
```
// <div class="default-action-button" role="button" tabindex="0">
.default-action-button:focus {
  background: green;
  outline: 3px solid blue;
}
```

Whether you move focus to a wrapper element with `tabindex="-1"` or a native button element is widely dependent on implementation. My hope is that `:focus-ring` would give us better control over additional scenarios beyond native buttons with custom CSS styles. But I think we're having difficulty reconciling all of these use cases at the same time, and the big question is whether we can reliably suppress focus for Slack's use case as well as for custom controls and focus management targets.

Here are the various use cases I would hope this polyfill could cover:

- Custom styles on natively focusable elements like `button:focus` that have historically showed persisted on click and triggered bug reports.
- Custom controls using ARIA and tabindex="0", both directly and indirectly focused, which persist both the default focus ring and custom CSS `:focus` styles on click.
- Focus management targets using tabindex="-1". (Should they not show focus at all? This is subjective)

This isn't an easy one to solve, so my apologies if this feels like piling on. But it was important for me to go back and remember my motivations for a tool like :focus-ring in the first place. Hopefully these details help bring that all into...focus. (I'll show myself out. :))

kloots:
@marcysutton the custom control issue you call out is a legit problem, *but* that's just a bug and (IMO) orthogonal to the issue of the heuristics for the keyboard behavior we should match on. I'll open an issue and PR to address that... one moment.

kloots:
@marcysutton actually, looking at the code and testing the demo we're already supporting custom controls save those with a `tabindex` set -1—which seems correct to me.

marcysutton:
@kloots how? When I click on a custom element with tabindex="0", the focus ring shows.  I want to make sure I didn't misunderstand the usage API.

kloots:
@marcysutton ohhhh that! Yes, **that is a bug.** In fact, I believe [I have fixed that in a separate PR](https://github.com/WICG/focus-ring/pull/37).

kloots:
@marcysutton I take it back, I cannot repro your bug in the [demo on master](https://wicg.github.io/focus-ring/demo/) or on this branch. Try clicking the `<div>` with a `tabindex` of 0 via the mouse. I don't see a focus ring except if I navigate to the element via the keyboard. Unless I still don't understand what you're talking about.

alice:
👍 

So sorry for letting this go cold!

robdodson:
@kloots do you think this is ready to merge? If so let's 🚀

kloots:
@robdodson It's ready to merge now. Just added one more commit from my fork which prevents handling `focus` and `blur` when the bubble to the `document` — which can caused errors in Firefox.

robdodson:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

robdodson:
I wanted to double check one last thing based on @marcysutton's comments.

In this example snippet you provided:

```
/* Hide focus ring by default */
:focus {
  outline: none
}

/* Show a focus outline when navigating via Tab */
:focus-ring {
  outline: 3px solid blue;
}

/* Ensure the default action button in dialogs has a focus outline */
.default-action-button:focus {
  background: green;
  outline: 3px solid blue;
}
```

The `.default-action-button` will always display a focus indicator of some kind. Including if it is clicked on with a mouse. Was that your intention? I think what you were trying to show was how to have it display a focus indicator in response to programmatic focus.

kloots:
@robdodson yes, expected. In that instance the context (if I remember) was a focus style for a button that mapped to the default action for a dialog, and as such, the focus style should always be visible regardless of how the button received focus: programmatic (on initial display of the dialog) or via user interaction (via keyboard or mouse).

robdodson:
@kloots ok. Just curious, do you have an example of using this behavior in Slack? I just wanted to see it in action if possible.

kloots:
@robdodson no example of that behavior in Slack (at least I don't think). But I _think_ this specific point is outside the scope of this PR and is the reason why @alice started a separate discussion for this in issue #42, no? Or is this blocking the merging of this PR?

robdodson:
I think it's blocking me because I'm starting to wonder if a one-size-fits-all :focus-ring style, by itself, is workable. Seeing Slack and Twitter fork it so they can add specific behavior is really interesting to me and makes me wonder if we need to add in the power of #42 at the same time. Otherwise it seems like someone will always be unhappy. As @marcysutton stated, I think she prefers the way it currently works and if we made this change she would switch back to WhatInput.

alice:
I say let's go ahead and merge (and apologies again for how long that took) - but definitely stay open to feedback from people who feel this change breaks things.

Issue #35, opened 2017-06-06
Remove support for author-added focus-ring class


kloots:
Filing up as a followup from [the discussion in issue #33 pertaining to the need for the `data-focus-ring-added` attribute](https://github.com/WICG/focus-ring/pull/34#pullrequestreview-42163894).  [As of 8d00580](https://github.com/WICG/focus-ring/commit/8d005806af5dc98b1ee2661ff764aaeda2150018) this polyfill supports persistence of instances of an author-defined `focus-ring` class. I would argue this should be removed on the premise that:

1. **This polyfill shouldn't give authors abilities they won't have with the actual :focus-ring psuedo.** Therefore, once actual support for :focus-ring lands in browsers the only change authors need to make to their CSS is changing their rules from .focus-ring to :focus-ring. Hence, more with the grain.

2. Should authors wish to ensure specific instances of UI always show focus they can leveraging the natural cascade and/or specificity rules in CSS:

```
// hide focus by default
:focus {
  outline: none;
}
// focus ring for keyboard navigation
.focus-ring {
  outline: solid 2px blue;
}

// always show focus outline around a specifc control
button.default-action:focus {
  outline: solid 2px blue;
}
```

alice:
The issue with this:

```js
// always show focus outline around a specifc control
button.default-action:focus {
  outline: solid 2px blue;
}
```

is that there is no way to recover the *default* focus style.

kloots:
@alice you didn't address my original point, which is once support for the actual `:focus-ring` pseudo lands in browsers, how would authors achieve the persistence you are currently offering in the polyfill? It doesn't seem like a good idea for this polyfill to temporarily give authors an ability they eventually won't have.

That aside, authors wishing to restore the browser's default focus style could apply a class to the html element identifying the UA, hide the focus ring by setting the `outline-width` to 0 and restore it by setting the `outline-width` to the correct default width per UA. Here's an example:

/* Hide focus ring by default */
:focus {
  outline-width: 0;
}

/* Show focus ring when navigating via keyboard */
.moz .focus-ring {
  outline-width: medium;
}

.chrome .focus-ring {
  outline-width: 5px;
}

/* Force focus ring to always be visible */
.moz .default-action:focus {
  outline-width: medium;
}

.chrome .default-action:focus {
  outline-width: 5px;
}
```

jonathantneal:
[...]

PR #36, opened 2017-06-06
Remove support for persisting author-defined focus-ring classes

kloots:
This PR addresses issue #35, removing support for persistence of instances of author-add `focus-ring` classes.

### DIFF
--- a/src/focus-ring.js
+++ b/src/focus-ring.js
@@ -48,7 +48,6 @@ document.addEventListener('DOMContentLoaded', function() {
     if (classList(el).contains('focus-ring'))
       return;
     classList(el).add('focus-ring');
-    el.setAttribute('data-focus-ring-added', '');
     // Keep a reference to the element to which the focus-ring class is applied
     // so the focus-ring class can be restored to it if the window regains
     // focus after being blurred.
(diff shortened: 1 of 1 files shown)


marcysutton:
I'm a little confused about how to use this polyfill. Can you also update the documentation for the current usage?

kloots:
@marcysutton [you'd simply install it as a project dependency via npm](https://github.com/WICG/focus-ring/blob/master/README.md#installation) and then include the built dist as part of your rollup.

marcysutton:
Sorry, I meant the selectors necessary to use the polyfill. I got a bit turned around by the discussion regarding CSS classes and how they should be avoided because developers will grow attached to them, and I assumed at first this change would have an impact on usage. 

In practicality, though, the `.focus-ring` class seems required until UAs actually implement `:focus-ring`.

jonathantneal:
I think going further in the direction of attributes is actually better than classes. I will pitch a PR, and I won’t be offended if it is rejected.

alice:
@marcysutton To try and clear up your understandable confusion:
This concerns an (admittedly poorly documented, that's on me) option to allow developer to opt in to *always* matching `.focus-ring` by simply adding a `.focus-ring` class.

This concept in general is something we need, and this seemed like a good option for the polyfill since we figured it would be something developer would attempt. It's an open question right now as to what the standards-based option for this might look like.

Issue #7, opened 2017-03-24
white list


WW3:
should [contenteditable], video[controls], audio[controls] also be white listed?

robdodson:
I believe items in whitelist are ones that should always display a focus ring.

[contenteditable] seems to, when I try it in Chrome, Firefox, and Safari.

audio and video controls don't seem to in Chrome or Safari but they do in Firefox.

My hunch is we wouldn't want to add those two because they work very similar to `<button>`. @alice wdyt?

bradkemper:
I think the general idea behind showing focus rings on text fields is that you can see where your typing is going to go if you start typing some words. You can then expect to see your words show up within that rectangle. So I would say [contenteditable] should be on the white list, but video controls should not be.

jonathantneal:
Depending on the result of #34, I can file a pull request to whitelist contenteditable.

[Test Page](https://s.codepen.io/jonneal/debug/b8cf0f3cff01f0a628d75f25cf9ac9d0)

I tested that page in **Edge** and **Internet Explorer 11**. None of these elements receive an outline from a pointer, but they all do from keyboard. In Chrome 48, Firefox 53, and Safari 10.1.1, `[contenteditable]` did receive outline from pointer, while media elements did not.

*As an aside, I follow @bradkemper’s rationale, and I think this behavior is unexpected.*

kloots:
@WW3 I've cut a PR to address this issue as the current behavior needs to be tuned IMO. Currently it requires an ARIA role of `textbox` to match and would also match if an element _only_ had a role of `textbox` and no `contenteditable` attribute—which isn't right. I think the right approach is simply keying off of the `contenteditable` attribute.

robdodson:
Looks like this is fixed by https://github.com/WICG/focus-ring/pull/37 so I'm going to close this issue.

PR #37, opened 2017-06-06
Ensure focus-ring class is applied to contenteditable elements

kloots:
This PR addresses issue #7 and ensures the `focus-ring` class is applied to contenteditable elements. Additionally, it removes the dom-matches polyfill dependency and updates the demo with more use cases.

### DIFF
--- a/dist/focus-ring.js
+++ b/dist/focus-ring.js
@@ -170,80 +170,27 @@ ClassList.prototype.toggle = function (token, force) {
   return (typeof force == 'boolean' ? force : !hasToken);
 };
 
-/**
- * Determine if a DOM element matches a CSS selector
- *
- * @param {Element} elem
- * @param {String} selector
- * @return {Boolean}
- * @api public
- */
-
-function matches(elem, selector) {
-  // Vendor-specific implementations of `Element.prototype.matches()`.
-  var proto = window.Element.prototype;
-  var nativeMatches = proto.matches ||
-      proto.mozMatchesSelector ||
-      proto.msMatchesSelector ||
-      proto.oMatchesSelector ||
-      proto.webkitMatchesSelector;
-
-  if (!elem || elem.nodeType !== 1) {
-    return false;
-  }
-
-  var parentElem = elem.parentNode;
-
-  // use native 'matches'
-  if (nativeMatches) {
-    return nativeMatches.call(elem, selector);
-  }
-
-  // native support for `matches` is missing and a fallback is required
-  var nodes = parentElem.querySelectorAll(selector);
-  var len = nodes.length;
-
-  for (var i = 0; i < len; i++) {
-    if (nodes[i] === elem) {
-      return true;
-    }
-  }
-
-  return false;
-}
-
-/**
- * Expose `matches`
- */
-
-var index$1 = matches;
-
 /* https://github.com/WICG/focus-ring */
 document.addEventListener('DOMContentLoaded', function() {
   var hadKeyboardEvent = false;
   var keyboardThrottleTimeoutID = 0;
   var elWithFocusRing;
 
-  // These elements should always have a focus ring drawn, because they are
-  // associated with switching to a keyboard modality.
-  var keyboardModalityWhitelist = [
-    'input:not([type])',
-    'input[type=text]',
-    'input[type=search]',
-    'input[type=url]',
-    'input[type=tel]',
-    'input[type=email]',
-    'input[type=password]',
-    'input[type=number]',
-    'input[type=date]',
-    'input[type=month]',
-    'input[type=week]',
-    'input[type=time]',
-    'input[type=datetime]',
-    'input[type=datetime-local]',
-    'textarea',
-    '[role=textbox]',
-  ].join(',');
+  var inputTypesWhitelist = {
+    'text': true,
+    'search': true,
+    'url': true,
+    'tel': true,
+    'email': true,
+    'password': true,
+    'number': true,
+    'date': true,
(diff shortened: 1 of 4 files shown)


src/focus-ring.js
```diff
@@ -36,7 +30,19 @@ document.addEventListener('DOMContentLoaded', function() {
    * @return {boolean}
    */
   function focusTriggersKeyboardModality(el) {
-    return matches(el, keyboardModalityWhitelist) && matches(el, ':not([readonly])');
+    var type = el.type;
+    var tagName = el.tagName.toLowerCase();
```
jonathantneal: You’re only testing `tagName` against `'input'` and `'textarea'`. Therefore, you could safely omit `.toLowerCase()` and test for `'INPUT'` and `'TEXTAREA'` instead.

src/focus-ring.js
```diff
@@ -74,11 +80,13 @@ document.addEventListener('DOMContentLoaded', function() {
    * opening a menu or dialog.
    */
   function onKeyDown() {
-    hadKeyboardEvent = true;
     // `activeElement` defaults to document.body if nothing focused,
     // so check the active element is actually focused.
-    if (matches(document.activeElement, ':focus'))
-      addFocusRingClass(document.activeElement);
+    var activeElement = document.activeElement;
+    if (activeElement.tagName.toLowerCase() == 'body')
```
jonathantneal: Similarly here, there is no browser that will return the tagName of `'BODY'` in lowercase.

kloots: @jonathantneal are you sure? I recall IE returning the declared case if in quirks mode. If you're 100% sure I'm happy to remove the use of `toLowerCase()`

jonathantneal: I am sure.

```html
<body></body><script>document.write(document.body.tagName)</script>
```

Screenshot of Internet Explorer 8 displaying the tagName of body in uppercase

In older IE, there would be an issue with new “unknown” elements (back then, HTML5 elements) returning lowercase. It’s unneeded here, because you’re testing for some of the oldest elements known to web. :)

src/focus-ring.js
```diff
@@ -31,7 +31,7 @@ document.addEventListener('DOMContentLoaded', function() {
    */
   function focusTriggersKeyboardModality(el) {
     var type = el.type;
-    var tagName = el.tagName.toLowerCase();
+    var tagName = el.tagName;
 
     if (tagName == 'input' && inputTypesWhitelist[type] && !el.readonly)
```
jonathantneal: Be sure to test these changes. I’m not sure `'input'` will work because it should now be uppercase?

dist/focus-ring.js
```diff
@@ -253,7 +200,19 @@ document.addEventListener('DOMContentLoaded', function() {
...
+    var type = el.type;
+    var tagName = el.tagName.toLowerCase();
+
+    if (tagName == 'input' && inputTypesWhitelist[type] && !el.readonly)
+      return true;
+
+    if (tagName == 'textarea' && !el.readonly)
+      return true;
+
+    if (el.contentEditable)
```
jonathantneal: The possible values of `contentEditable` are `'inherit'`, `'true'`, or `'false'`.

kloots: Fixed.

src/focus-ring.js
kloots: Fixex

kloots:
@alice, any thoughts on this PR?

src/focus-ring.js
```diff
@@ -74,11 +80,13 @@ document.addEventListener('DOMContentLoaded', function() {
...
-    hadKeyboardEvent = true;
     // `activeElement` defaults to document.body if nothing focused,
     // so check the active element is actually focused.
-    if (matches(document.activeElement, ':focus'))
-      addFocusRingClass(document.activeElement);
+    var activeElement = document.activeElement;
+    if (activeElement.tagName == 'BODY')
+      return;
+
+    hadKeyboardEvent = true;
```
alice: Why was this moved below the check?

kloots: Because the more I thought about it, we don't want to set that to true if focus hasn't moved to a focusable control. This way we can use `hadKeyboardEvent` to determine if a focus moved to a focusable element as a result of keyboard navigation.

alice: `hasKeyboardEvent` is not intended for that use, though. I think this could cause a bug if a keyboard event triggered a focus move (for example, activating a button which caused a piece of modal UI to appear). Please revert this change.

kloots: Sorry, I'm not 100% following you regarding the bug that could result from this change. Mind explaining in great detail?

alice: Ah - I had it backwards. However, since the keyboard event precedes the focus event, this is checking where focus is coming _from_, rather than going _to_, isn't it?

kloots: No worries at all. To answer your question: it's the other way around. The fact the keyboard event precedes the focus event actually works to our advantage here because we know the event sequence is: keydown then focus, and therefore the focus handler can key off of `hadKeyboardEvent` to know focus happened as a result of keyboard, and then add the focus ring class.

alice: Right: so at the point where this fires, `activeElement` is the element which is about to _lose_ focus, isn't it? So why _not_ set `hadKeyboardEvent` based on that?

kloots: Ah, yes! Now I see what you're saying. Will change.

alice:
👍

Issue #42, opened 2017-06-18
Brainstorm: options for opting in to always matching focus ring


alice:
We know we need a way to opt an element in to always (or never?) getting a focus ring. Let's talk about strategies for doing this.

alice:
**Proposal: just use `:focus`**

Use the existing `:focus` pseudo-selector to opt an element in to always getting a focus ring on focus. Requires a way to recover the platform default focus style, assuming `:focus-ring` will be the UA default in future.

**Example**

```css
:focus {
  outline: platform-default;  /* need a way to express this */ 
}
```

alice:
**Proposal: add a new CSS property to express when an element should match `:focus-ring`**

Allow an element to opt in to `:focus-ring` matching the way built-in text editing elements do. Potentially also allows fine tuning focus-ring matching at page author discretion.

**Example**
/* all syntax to be considered a "straw man" */

```css
custom-text-editor {
  show-focusring: always;
}

div[tabindex=-1] {
  show-focusring: never;  /* may potentially be overridden by user preference */
}

* {
  /* match when focus is moved via the keyboard explicitly, i.e. tab/shift-tab */
  show-focusring: explicit-focus; 
}

button {
  show-focusring: on-keypress;
}
```

alice:
cc @shans

kloots:
@alice I think your 2nd proposal is far far better than pursuing the current `:focus-ring` implementation; it's more robust. 

1. Allows for fine grain control over when the focusing ring is displayed; solving for all of the various niche cases discussed in #33
2. Makes it much easier for authors to preserve browser default focus styles (should they want to)
3. Feels more with-the-grain of CSS; reminds me of how `overflow` works.

Nice one.

robdodson:
I like the second proposal if we can get buy-in. And I also want `outline: platform-default;` :)

kloots:
@alice @robdodson what's the next step for moving on @alice's proposal?

robdodson:
@kloots I think one thing alice and I were unsure of: were you saying you like the second proposal _instead of `:focus-ring`_ or in addition to `:focus-ring`? In other words, were you suggesting we drop `:focus-ring` entirely and pursue that other proposal?

kloots:
@robdodson yeah, I _really_ like @alice's second proposal and think it'd be better for the industry if we pursued that over `:focus-ring`.

alice:
[...]

PR #43, opened 2017-06-18
Revert #36

alice:
On reflection, I'm not comfortable with stomping author-provided CSS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,9 @@ by selecting for the case when `.focus-ring` is _not_ applied:
 }
 ```
 
-Note that `.focus-ring` will *always* be removed on blur,
-so authors should not explicitly add this class.
+If there are elements which should always have a focus ring shown,
+authors may explicitly add the `focus-ring` class.
+If explicitly added, it will not be removed on `blur`.
 
 ## Rationale
 
--- a/src/focus-ring.js
+++ b/src/focus-ring.js
@@ -54,6 +54,7 @@ document.addEventListener('DOMContentLoaded', function() {
     if (classList(el).contains('focus-ring'))
       return;
     classList(el).add('focus-ring');
+    el.setAttribute('data-focus-ring-added', '');
     // Keep a reference to the element to which the focus-ring class is applied
     // so the focus-ring class can be restored to it if the window regains
     // focus after being blurred.
@@ -66,7 +67,10 @@ document.addEventListener('DOMContentLoaded', function() {
    * @param {Element} el
    */
   function removeFocusRingClass(el) {
+    if (!el.hasAttribute('data-focus-ring-added'))
+      return;
     classList(el).remove('focus-ring');
+    el.removeAttribute('data-focus-ring-added');
   }
 
   /**


alice:
@robdodson PTAL?

@kloots PTAL/FYI

robdodson:
My 2 cents, I think for now we should remove the `.focus-ring` class on blur because the currently specced `:focus-ring` pseudo would also remove on blur.

kloots:
@alice what's your logic on this revert? Don't we want the behavior of this polyfill to line up as closely as possible with how the eventual `:focus-ring` pseudo class will work? And in that case authors would be responsible for leveraging the cascade / specificity rules for managing what elements should always have a focus outline.

kloots:
Also, I just had to Google "PTAL"... :)

alice:
The logic is that if the user has *added* a `focus-ring` class, for whatever reason, without this revert our code stomps it and removes it on blur. I'm not comfortable stomping *author-added* CSS like this.

alice:
(Also, PTAL :smile: at https://github.com/WICG/focus-ring/issues/42 for the discussion of how to allow authors to opt in to focus rings)

alice:
Here's an example of how the current code could cause weird behaviour:
- Author doesn't read the README carefully (nobody ever does), sees the advice to key styling off `.focus-ring` class, has an element which should always get a focus ring.
- Author thinks "cool, I'll just add the class manually for that thing and it'll match the style I set up", and does so.
- First time they click the element, it gets a focus ring. Awesome, it works!
- Second time they click it, after focus moves away, it doesn't? Wat?

We can't get complete parity with the pseudo-selector in the polyfill, no matter what, so theoretical purity is not a priority for me. I would rather not enforce "correct" usage by authors, but would prefer to make the fewest assumptions possible about how people will choose to use it.

Once folks *can* use the pseudo-selector, they can still use a `.focus-ring` class to force focus to apply if they want to do it that way, with or without some other mechanism (see #42).

kloots:
@alice thanks for the explanation. I understand your caution. Rather than account for this edge case now, can we instead:

1. Update the README with guidance to authors wishing to force a focus outline on certain elements
2. Wait to see how many authors encounter this case _and_ file issues

This way we're able to:
1. Get a much clearer sense of actual implementation problems encounter by developers
2. Determine the best course of action based on the volume and severity of point number 1

alice:
I'm definitely happy to pull an update to the README with more consistent advice, but I feel like the rest of what you're proposing is leaving a trap for developers to see how many find it, which I'm not keen on.

kloots:
Sorry, perhaps I wasn't explaining my reasoning well enough: all I'm suggesting is we consider fixing this issue if it actually comes up as an issue. Right now you're keeping this in on a hunch.

robdodson:
Just to hop in again.

Frankly I think you're both "right" and this is kind of an edge case so it's possible we'll never get any issues related to it no matter which solution we go with 😄 

As I've been going back and forth on this, I'm now leaning toward reverting the change because of the use case Alice cited. Also if I use the polyfill and manually add `.focus-ring` classes, and down the line `:focus-ring` finally lands, then I can leave the `foo.focus-ring` styles in my CSS / JS and I only need to switch `:focus:not(.focus-ring)` to `:focus:not(:focus-ring)`.

kloots:
@robdodson re:

> this is kind of an edge case so it's possible we'll never get any issues related to it no matter which solution we go with

Fixing an issue that we don't yet have (or will possibly ever have) doesn't seem like a good reason for reverting. :)

robdodson:
> Fixing an issue that we don't yet have (or will possibly ever have) doesn't seem like a good reason for reverting. :)

Just trying to be constructive. From my perspective the issue seems like a bit of a bikeshed but we're expending a lot of energy on it. I'd like to be the tie breaker vote so we can move forward.

I think Alice's point about folks not reading the docs is consistent with my own tendency to not RTFM 😸 so I'm sympathetic to that angle and not wanting to surprise the user.

kloots:
Bikeshed for sure. Also in the interest of moving on: that's why I suggested simply deferring this decision out until we get actual user feedback or bug reports. That way no one has to be the tie breaker.

bkardell:
I actually agree with this revert - I understand what @kloots is saying but the reality is that this makes things pretty easy and being able to 'plug into' this is something that was discussed and I don't want to say that anything CSSWG has written down is final or in blood until it's actually implemented - that topic already has other discussions in CSSWG and helping to figure out and pave a path seems like... pretty good to me, and better this way.

alice:
Ok, we have three in favour of this revert so I'm going to go ahead and revert it :)

PR #45, opened 2017-07-13
Fix: Async script not init polyfill

Fyzu:
If the script is loaded asynchronously, the `DOMContentLoaded` event will not be called, therefore, the script will not be initialized.

image

[More information here](https://www.kirupa.com/html5/running_your_code_at_the_right_time.htm)


robdodson:
Thanks for sending this in. Will try to review in the morning

alice:

PR #50, opened 2017-08-10
Add body class when loaded - fix #8

### LINKED ISSUE #8: Global :focus { outline: none; } reset declaration and Polyfill concerns
Hi :)

Per the documentation, I'm really struggling with the blanket `:focus { outline: none; }` code recommendation. I want to be on board with this awesome feature but feel like we could come up with something better.

It is very bold to say hey.. attention all browsers - kill all outlines on focusable elements and then rely on JS to fill the gaps or when implemented only supporting browsers will overwrite that rule. Many older browsers support `:focus {}` but will not support `:focus-ring` ever.

## Some ideas

Here are some ideas instead of just a global `:focus { outline: none; }`

1. For the spec: lean on fault tolerance so browsers that don't understand skip it?:

`:not(:focus-ring) :focus { outline: none; }`

2. For the Polyfill have JS add a class to `html` tag and descend from that:

`.focus-ring-polyfill :focus:not(.focus-ring) { outline: none; }`

This would mean JS is probably present, loaded, no errors, and the Polyfill should work.

I don't want to be a stick in the mud but we are working hard out here trying to get folks to stop killing outlines for many different accessibility-related reasons.

## Consider these personas:

**Lisa:** I use my Dell laptop I'm a mouse user, touch user, and a keyboard user. I have a cognitive related disability that affects my memory and have low-vision it's important for me to have an understanding of where I am in the interface -- and yes even right after I click on something. I don't always use a screen reader. I may start a web form using a keyboard and then click on something and I get confused when the focus ring is removed from the UI.

**Paul:** I turn JavaScript off because it ensures my browsing experience does not have too much animation. I'm epileptic, have low vision, and very sensitive to animation and flashing. I'm also a keyboard user and mouse user. Because of this I turn the brightness down on my screen. I have difficulty getting around a site when they don't have focus states.

stefanjudis:
I was just about to implement the polyfill and came across #8. It addresses my concerns completely because the CSS I have in my project is as follows. 👇 

```
  :focus {
    outline-style: dashed;

    &:not(.focus-ring) {
      outline: none;
    }
  }
```

This can destroy the complete focus handling in case of JS failure or loading troubles.

This PR adds the discussed behavior of the polyfill to add a `focus-ring-enabled` class to the body element when it's done doing its magic. ;)

### DIFF
--- a/dist/focus-ring.js
+++ b/dist/focus-ring.js
@@ -288,15 +288,15 @@ function init() {
    * to which it was previously applied.
    */
   function onWindowFocus() {
-    if (document.activeElement == elWithFocusRing) {
+    if (document.activeElement == elWithFocusRing)
       addFocusRingClass(elWithFocusRing);
-      elWithFocusRing = null;
-    }
+
+    elWithFocusRing = null;
   }
 
   /**
-   * When the window regains focus, restore the focus-ring class to the element
-   * to which it was previously applied.
+   * When switching windows, keep track of the focused element if it has a
+   * focus-ring class.
    */
   function onWindowBlur() {
     if (document.activeElement.classList.contains('focus-ring')) {
@@ -312,6 +312,8 @@ function init() {
   document.addEventListener('blur', onBlur, true);
   window.addEventListener('focus', onWindowFocus, true);
   window.addEventListener('blur', onWindowBlur, true);
+
+  document.body.classList.add('focus-ring-enabled');
 }
 
 /**
--- a/README.md
+++ b/README.md
@@ -16,10 +16,11 @@ in situations in which the `:focus-ring` pseudo-selector should match.
 
 We suggest that users
 selectively disable the default focus style
-by selecting for the case when `.focus-ring` is _not_ applied:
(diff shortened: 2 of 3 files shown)


stefanjudis:

dist/focus-ring.js
```diff
@@ -288,15 +288,15 @@ function init() {
...
+
+    elWithFocusRing = null;
   }
 
   /**
-   * When the window regains focus, restore the focus-ring class to the element
-   * to which it was previously applied.
+   * When switching windows, keep track of the focused element if it has a
+   * focus-ring class.
    */
```
stefanjudis: These JSDOC changes are not intended by me. It seems like the docs in the dist file were not up to date. Now they're matching with the docs in `src`.

robdodson:
LGTM thank you @stefanjudis!

stefanjudis:
Thanks so much for merging.

Do you plan to also release a new version? 😊

robdodson:
Yep I just want to check in with @alice first

Issue #54, opened 2017-09-12
Focus ring does not account for radio buttons


johnevanofski:
I noticed that the focus-ring polyfill only looks for tab key presses, and not arrow input, which is part of the default behavior for radio buttons. 

I've fixed this locally by adding a few lines to the `onKeyDown` function.

```
function onKeyDown(e) {
    if (e.altKey || e.ctrlKey || e.metaKey)
      return;

    if (
	e.keyCode != 9 // tab
	&& e.keyCode != 37 // left arrow
	&& e.keyCode != 38 // up arrow
	&& e.keyCode != 39 // right arrow
	&& e.keyCode != 40 // down arrow
	)
      return;

    hadKeyboardEvent = true;
  }
```

fstorr:
@johnevanofski Thanks for this—it's just what I needed.

robdodson:
@alice can you think of any reason why we wouldn't want to add this to the library? Otherwise I think we should PR it

robdodson:
@alice friendly ping on this. are you cool with me doing a PR or can you think of a reason we wouldn't want to include this?

majornista:
@robdodson I use the following to evaluate `event.key`, that way the polyfill will work for tablists, list boxes, radio buttons, or a dialog closed with Escape key or keyboard activation of an OK button:

```js
  var NAVIGATION_KEYS = [
    'Tab',
    'ArrowUp',
    'ArrowRight',
    'ArrowDown',
    'ArrowLeft',
    'Home',
    'End',
    'PageUp',
    'PageDown',
    'Enter',
    ' ',
    'Escape',

    /* IE9 and Firefox < 37 */
    'Up',
    'Right',
    'Down',
    'Left',
    'Esc'
  ];

  function onKeyDown(e) {
    if (e.altKey || e.ctrlKey || e.metaKey || NAVIGATION_KEYS.indexOf(event.key) === -1)
      return;

    hadKeyboardEvent = true;
  }
```

robdodson:
@kloots I'm curious what your thoughts are on this? Do you think expanding the keys that trigger keyboard modality to the above set will affect your keyboard shortcut use case?

kloots:
@robdodson I think @johnevanofski is right, we _should_ account for this behavior as browsers do move focus between radio buttons in a group (radio buttons with the same `name`). So, the `onKeyDown` listener should first check if the target of the event is a radio button before responding to an arrow key press.

majornista:
[...]

PR #57, opened 2017-10-06
Issue #54: Focus ring does not account for radio buttons

majornista:
Per TODO in documentation:

> TODO: ideally, we also trigger keyboard modality following a keyboard
event which activates an element or causes a mutation; this still needs
to be implemented.

Include other types of keyboard input than the tab key that may produce
focus navigation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,4 @@ The script uses two heuristics to determine whether the keyboard is being used:
 or `Shift + Tab`.
 - focus moves into an element which requires keyboard interaction,
   such as a text field
-- _TODO: ideally, we also trigger keyboard modality
-  following a keyboard event which activates an element or causes a mutation;
-  this still needs to be implemented._
+- We also trigger keyboard modality following a keyboard event which activates an element or causes a mutation. The following keys that typically result in a change of context or focus are handled: `Backspace`, `Enter`, `Esc`, `Space`, `PageUp`, `PageDown`, `End`, `Home`, `ArrowLeft`, `ArrowUp`, `ArrowRight`, `ArrowDown`, or `Delete`.
--- a/dist/focus-ring.js
+++ b/dist/focus-ring.js
@@ -1,7 +1,7 @@
 (function (global, factory) {
-	typeof exports === 'object' && typeof module !== 'undefined' ? factory() :
-	typeof define === 'function' && define.amd ? define(factory) :
-	(factory());
+  typeof exports === 'object' && typeof module !== 'undefined' ? factory() :
+  typeof define === 'function' && define.amd ? define(factory) :
+  (factory());
 }(this, (function () { 'use strict';
 
 /**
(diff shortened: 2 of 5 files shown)


robdodson:
@marcoscaceres This PR is flagged as having IPR issues but I'm not sure when it's ok to "Mark as non-substantive". Are there rules that help define what is and isn't ok?

marcoscaceres:
> @marcoscaceres This PR is flagged as having IPR issues but I'm not sure when it's ok to "Mark as non-substantive". Are there rules that help define what is and isn't ok?

The rule is: typos, grammar, rephrasing sentences, fixing up or adding examples/diagrams/notes, are all non-substantive. Anything that affects functionality/behavior or changes and algorithm is substantive.   

If you are unsure, or it's unclear, then it's safer to say it's substantive and ask the contributor to join the WICG. And as always, if you are still unsure - just ping the chairs for review.

robdodson:
ok so @majornista will need to visit this page https://www.w3.org/community/wicg/, read through it, and click 'Join Group', is that correct @marcoscaceres ?

marcoscaceres:
>  is that correct @marcoscaceres ?

Yep, that is correct. Sorry for the additional pain.

majornista:
Thanks, @robdodson and @marcoscaceres. I've joined the group, so we should now be able to discuss this PR more substantively.

srinivasvenigalla:
Can you please add "button" to the "selectable":"role" please?

robdodson:
This should be addressed by #68

srinivasvenigalla:
Hi, could you please add "menuitem" to selectables? Menuitems are also clickable..

robdodson:
@srinivasvenigalla this PR was closed by #68. The polyfill does not special case ARIA roles because it's meant to match the browser heuristic and that does not take ARIA into account.

For `menuitem` are you expecting that it should match `:focus-ring`? If so, I think what you actually want is https://github.com/WICG/focus-ring/issues/42#issuecomment-309309323 which is an additional CSS property we're proposing that will let you hint to the browser when `:focus-ring` should match.

Until we add that property to the polyfill, the easiest thing to do is [manually add the `.focus-ring` class yourself](https://github.com/WICG/focus-ring#2-update-your-css).

> If there are elements which should always have a focus ring shown, authors may explicitly add the focus-ring class. If explicitly added, it will not be removed on blur.

srinivasvenigalla:
@robdodson if we add .focus-ring manually, it would display the focus ring for mouse also, which is not our UX would approve.

robdodson:
@srinivasvenigalla sorry my mistake.

Thinking about this a bit more, I think what you're requesting should already work with the latest version because it sets keyboard modality if the user uses the arrow keys. https://github.com/WICG/focus-ring/blob/master/src/focus-ring.js#L82-L87

So if your menuitems are just div's with tabindex=0 and the user presses the up/down arrow keys to move between them, that should apply .focus-ring.

Can you verify you're using the latest version?

PR #58, opened 2017-10-17
Fix missing activeElement. Fixes #56

### LINKED ISSUE #56: Exception: A DOM Element reference is required
I just put this into production on unsplash.com and noticed that we're getting thousands of client-side exceptions reported due to `A DOM Element reference is required`. The offending line is here https://github.com/WICG/focus-ring/blob/0e9f086bbc59ca84a2a932e89df0b71bce18b643/dist/focus-ring.js#L25

I haven't been able to reproduce this locally. Do you have any ideas why this might be happening? Have you seen it before?

Thanks!

robdodson:


robdodson:
@alice FYI

Issue #15, opened 2017-03-29
No focus styling when navigating with keyboard from outside the web content


erikkroes:
Tested on https://wicg.github.io/focus-ring/demo/

Using safari: when the address bar has the focus, pressing tab moves focus to the first input field but it does not receive focus styling. Pressing alt+tab moves focus to the last input field but it does not receive focus styling.

Using chrome: when the address bar has the focus, pressing alt+tab moves focus to the last input field but it does not receive focus styling.

In both cases, if I press shift after navigating, it does add the styling. I assume there's a problem with catching the keyboard events from outside the web content?

robdodson:
hm that's interesting. In the Safari example, since it's just input[type=text] I would imagine this line https://github.com/WICG/focus-ring/blob/gh-pages/src/focus-ring.js#L103 would kick in to apply the focus-ring class since the target should be in the whitelist. Will take a look later today. Thanks for the ticket!

patrickhlauke:
My guess would be that the keyboard event is "swallowed"/consumed by the actual browser UI, so does not trigger the `onKeyDown` and therefore does not set the `hadKeyboardEvent`. don't think there's any way to catch this until the user actually does a keystroke once in the actual web content

patrickhlauke:
actually, in the case of Chrome and reverse-tabbing to the last element in the page - as it's an `<input type="file">` - this does apply. the keyboard event happened in the UI, so `hadKeyboardEvent` is `false` at that point, and a `file` type of input is not in the whitelist ... so that's why when SHIFT+TABbing from address bar to that file chooser the focus is not applied.

kloots:
I cannot reproduce this bug in Safari as reported. I'm testing in Safari Version 10.1 (12603.1.30.0.34) on Mac OS 10.12.4 (16E195).

kloots:
Update: I _was_ able to reproduce this providing the first element in the page is **not** a text input. Easily reproducible in both Safari and Chrome if you change [the demo](https://wicg.github.io/focus-ring/demo/) such that a `<button>` is the first focusable element in the page.

kloots:
Some additional findings: this bug is exclusive to Safari and Chrome; IE and Firefox don't have this issue. The reason: neither Chrome or Safari fire a `keydown` event in advance of focus moving from the browser's location bar into the document. As a result, there's no means of determining if the initial`focus` event was preceded by a `keydown` event.

[...]

PR #60, opened 2017-10-23
Style keyboard focus on keyup

kloots:
This PR fixes issues #15 **and** #54. The core idea is to changeup _how_ we style focus, moving from adding the `focus-ring` class in response to `focus` events preceded by `keydown` events, to simply leveraging `keydown` instead.

As noted in the code comments: `keyup` provides two advantages over our current approach:
- It will reliably fire when the user is navigating from the browser chrome into the document (fixing issue #33).
- Requires one less event listener, while still ensuring we add `focus-ring` only in response to `focus` originating from the keyboard.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -37,6 +37,20 @@
       You should see the normal platform focus indicator on elements below differently based on whether you interact with them via keyboard or mouse.
       If you are using the mouse and get lost, at any time you can press a key (such as Shift) to have the focus indicator appear.
     </p>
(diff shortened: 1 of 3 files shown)


alice:
This is a great fix, but looks like the pull is against an older version of master - looks like you need to sync your workspace?

kloots:
@alice so sorry about that! Should be all tidied up now!

alice:
:+1:

PR #63, opened 2017-11-02
Add focus-ring class in response to arrow keys for radio buttons

kloots:
This PR fixes issue #54 by adding adding/removing the `focus-ring` class when the event target is a radio button.


robdodson:
I think this is fixed by #68

PR #64, opened 2017-11-02
Add support for styling focus for ARIA controls requiring managed focus

kloots:
This PR is a proof of concept and aims to extend `focus-ring` support to [ARIA controls requiring managed focus](https://www.w3.org/TR/wai-aria/roles#managingfocus).

**Goals for this approach were**
- *Opt in*: only extend this support to [ARIA controls requiring managed focus](https://www.w3.org/TR/wai-aria/roles#managingfocus); existing heuristics with respect to when the `focus-ring` class is applied to native HTML controls remain unchanged.
- *Scalable*: support any keyboard or programmatic focus styling as required by the developer implementing the control *AND* support both types of focus management: managed tabIndex or use of [`aria-activedescendant`](https://www.w3.org/TR/wai-aria/roles#aria-activedescendant)
- *With-the-grain*: shouldn't introduce any new or non-standard attributes or practices

### DIFF
--- a/dist/focus-ring.js
+++ b/dist/focus-ring.js
@@ -197,6 +197,13 @@ function init() {
     'datetime-local': true,
   };
 
+  var arrowKeys = {
+    37: true,
+    38: true,
+    39: true,
+    40: true,
+  };
+
   /**
    * Computes whether the given element should automatically trigger the
    * `focus-ring` class being added, i.e. whether it should always match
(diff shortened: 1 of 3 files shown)


dist/focus-ring.js
```diff
@@ -244,6 +273,35 @@ function init() {
...
   }
 
+  /**
+   * Validate the key we're handling should result in the application of the
+   * `focus-ring` class.
+   * @param {Event} e
+   * @return {Boolean}
+   */
+  function keyIsValid(e) {
+    var keyCode = e.keyCode;
```
stephan281094: `e.keyCode` is deprecated. Use `e.code` instead.
See https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode.

robdodson:
It looks like KeyboardEvent.code is not supported in Internet Explorer,
according to the MDN support tables.

dist/focus-ring.js
stephan281094: Never mind. As Rob pointed out, `KeyboardEvent.code` is not supported in Internet Explorer.
Thanks for the heads up @robdodson.

robdodson:
I think I'm going to close this, though I definitely appreciate the work put into it.

My thinking on this is that any control, if interacted with using a keyboard, should probably show focus states. I can't off the top of my head think of one where you wouldn't want this?

For keyboard shortcuts that move focus to an interactive control, you probably want a `:focus-visible` to match on that control. Otherwise you can end up with a very mysterious focus state as this gif demonstrates (mouse clicking the button moves focus to the other control which is activated with spacebar, but there's no clear indicator that this is happening).

click

For keyboard shortcuts that move focus to a new region in the page, like into a sidebar or modal window, I think you want a [focus navigation start point API](https://developers.google.com/web/updates/2016/03/focus-start-point) (today browsers do this but developers can't call it). This would let you avoid the annoying tabindex=-1 trick just to move the focus start point, and you wouldn't need to keep applying `outline: none` to those non-interactive regions. My hope is over the next quarters we can start to spec out and polyfill focus navigation start point.

Issue #44, opened 2017-07-12
Suggestion: consider offering a distribution without baked-in DOM polyfills


emilbjorklund:
One thing currently prohibiting me from using this (very useful) library is that it bakes in polyfills for both `classList` and `matches` (as of v1.0.1) - both of which I already polyfill elsewhere in many of my projects.

Is there any possibility of offering one file in the `dist/` directory that does not include these?

emilbjorklund:
Ah - just saw now that the `classList` lib is not a polyfill per se - which makes it a bit trickier...

robdodson:
I think we removed `matches` in the source.

Does the inclusion of the `classList` polyfill break anything for you, or is this more of a file size concern?

emilbjorklund:
> I think we removed matches in the source.
Yup – I caught that after posting, hence the edit. 😊

> Does the inclusion of the classList polyfill break anything for you, or is this more of a file size concern?

More of a filesize concern, I guess, but also some sort of principle. It's really nice building modern DOM stuff and just polyfilling as needed (and I’m really glad we need less polyfills/libs these days for the simple stuff) but it kinda defeats the purpose if everyone’s polyfilling or including small dependencies to fix minor browser inconsistencies in their own lib, if that makes sense?

Anyway, completely understand if it’s not important at the moment. I notice it’s only used in three places though, so one way of getting rid of the dependency would be to just write it with some old-school `className` manipulation. 😜

robdodson:
It's not that it's not important, I'm just not sure what the best approach is. I guess we could tell folks what we depend on and tell them they need to feature detect and manually load those polyfills themselves? Are there better alternatives that you've seen?

robdodson:
Thinking through this a bit more...

We could tell folks to use polyfill.io to load the necessary polyfills.
Or we could distribute two versions. A polyfilled one and a non-polyfilled one. And let folks decide.

emilbjorklund:
@robdodson Sorry for the radio silence! Re: your previous comment, I personally tend to like the approach of two builds (polyfilled vs non-polyfilled) in general.

robdodson:
Fixed on master.

PR #68, opened 2017-11-15
Revert behavior to fix failing tests. Add arrow key support. Fixes #65, #54, #63, #44

### LINKED ISSUE #65: Wrong behaviour according to CSS editor drafts?
According to:
https://drafts.csswg.org/selectors-4/#the-focusring-pseudo

The pseudo class focus-ring should always be applied on focus to input elements to indicate a where the user should enter text.
However, it should applied if I select e.g. a link / button element using the keyboard (tab button). 
However, if I do so with the mouse it shouldn't be applied. The reason is that if I tab through links using the keyboard I need to know WHERE I am but I don't need that information if I click on the element.
Of course, it depends on the corresponding browser's heuristics but that part is mentioned as example there.

robdodson:
This reverts the behavior that's currently on master because it was breaking mouse focus and causing most of the tests to fail. It uses a different approach to address the URL bar issue from #15 and adds support for arrow keys to address #54, #63.

This also removes the polyfill for classList. This is in accordance with the W3C's new [Polyfill guidance](https://www.w3.org/2001/tag/doc/polyfills/#don-t-serve-unnecessary-polyfills) which suggests that polyfills should not bundle other polyfills. Instead, we'll update our docs to mention that the polyfill depends on classList and users should use a service like polyfill.io to add it.

@kloots FYI because I needed to back out the previous `onKeyUp` change.

### DIFF
--- a/src/focus-ring.js
+++ b/src/focus-ring.js
@@ -1,17 +1,12 @@
-import classList from 'dom-classlist';
 
 /**
  * https://github.com/WICG/focus-ring
  */
 function init() {
+  var hadKeyboardEvent = true;
   var elWithFocusRing;
 
   var inputTypesWhitelist = {
-    'radio': true,
-    'checkbox': true,
-    'button': true,
-    'reset': true,
-    'submit': true,
     'text': true,
     'search': true,
     'url': true,
@@ -56,9 +51,9 @@ function init() {
    * @param {Element} el
    */
   function addFocusRingClass(el) {
-    if (classList(el).contains('focus-ring'))
+    if (el.classList.contains('focus-ring'))
       return;
-    classList(el).add('focus-ring');
+    el.classList.add('focus-ring');
     el.setAttribute('data-focus-ring-added', '');
   }
 
@@ -70,32 +65,39 @@ function init() {
   function removeFocusRingClass(el) {
     if (!el.hasAttribute('data-focus-ring-added'))
       return;
-    classList(el).remove('focus-ring');
+    el.classList.remove('focus-ring');
     el.removeAttribute('data-focus-ring-added');
   }
 
   /**
-   * On `keyup` add `focus-ring` class if the user pressed Tab and the event
-   * target is an element that will likely require interaction via the
-   * keyboard (e.g. a text box).
-   * The `keyup` event is used over the focus event because:
-   * 1. `focus` is a device-independent event, and `keyup` ensures the
-   *    `focus-ring` class is only added when focus originates from
-   *    keyboard navigation.
-   * 2. Unlike `focus`, keyup` will fire when the user navigates from the
-   *    browser chrome into the document. (For more, see issue #15)
+   * On `keydown`, set `hadKeyboardEvent`, add `focus-ring` class if the
+   * key was Tab.
    * @param {Event} e
    */
-  function onKeyUp(e) {
+  function onKeyDown(e) {
     if (e.altKey || e.ctrlKey || e.metaKey)
       return;
 
     if (e.keyCode != 9)
       return;
 
-    var target = e.target;
-    if (focusTriggersKeyboardModality(target)) {
-      addFocusRingClass(target);
+    hadKeyboardEvent = true;
+  }
+
+  /**
+   * On `focus`, add the `focus-ring` class to the target if:
+   * - the target received focus as a result of keyboard navigation
+   * - the event target is an element that will likely require interaction
+   *   via the keyboard (e.g. a text box)
+   * @param {Event} e
+   */
+  function onFocus(e) {
+    if (e.target == document)
+      return;
+
+    if (hadKeyboardEvent || focusTriggersKeyboardModality(e.target)) {
+      addFocusRingClass(e.target);
+      hadKeyboardEvent = false;
     }
   }
 
@@ -115,10 +117,7 @@ function init() {
    * to which it was previously applied.
    */
   function onWindowFocus() {
-    // When removing the activeElement from DOM it's possible IE11 is in state
-    // document.activeElement === null
-    if (!document.activeElement)
-      return;
+    window.removeEventListener('focus', onWindowFocus, true);
     if (document.activeElement == elWithFocusRing)
       addFocusRingClass(elWithFocusRing);
 
@@ -128,13 +127,15 @@ function init() {
   /**
    * When switching windows, keep track of the focused element if it has a
    * focus-ring class.
+   * @param {Event} e
    */
-  function onWindowBlur() {
-    // When removing the activeElement from DOM it's possible IE11 is in state
-    // document.activeElement === null
-    if (!document.activeElement)
+  function onWindowBlur(e) {
+    if (e.target !== window)
       return;
-    if (classList(document.activeElement).contains('focus-ring')) {
+
+    window.addEventListener('focus', onWindowFocus, true);
+    addInitialPointerMoveListeners();
+    if (document.activeElement.classList.contains('focus-ring')) {
       // Keep a reference to the element to which the focus-ring class is applied
       // so the focus-ring class can be restored to it if the window regains
       // focus after being blurred.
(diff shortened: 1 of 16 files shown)


src/focus-ring.js
```diff
@@ -70,32 +65,43 @@ function init() {
...
-   *    browser chrome into the document. (For more, see issue #15)
+   * On `keydown`, set `hadKeyboardEvent`, add `focus-ring` class if the
+   * key was Tab/Shift-Tab or Arrow Keys.
    * @param {Event} e
    */
-  function onKeyUp(e) {
+  function onKeyDown(e) {
+    const allowedKeys = [9, 37, 38, 39, 40];
+
+    // If the user is holding down a modifier key, abort.
```
alice: Suggestion: "Ignore keypresses if the user is holding down a modifier key"

src/focus-ring.js
```diff
@@ -70,32 +65,43 @@ function init() {
...
     if (e.altKey || e.ctrlKey || e.metaKey)
       return;
 
-    if (e.keyCode != 9)
+    // If the key can't be found in the list of allowed keys, abort.
```
alice: Suggestion: "Ignore keypresses which aren't related to keyboard navigation."

src/focus-ring.js
```diff
@@ -70,32 +65,43 @@ function init() {
...
+    if (allowedKeys.indexOf(e.keyCode) === -1)
+      return;
+
+    hadKeyboardEvent = true;
+  }
+
+  /**
+   * On `focus`, add the `focus-ring` class to the target if:
+   * - the target received focus as a result of keyboard navigation
```
alice: s/navigation/navigation, or/

src/focus-ring.js
```diff
@@ -128,26 +131,75 @@ function init() {
...
       // Keep a reference to the element to which the focus-ring class is applied
       // so the focus-ring class can be restored to it if the window regains
       // focus after being blurred.
       elWithFocusRing = document.activeElement;
     }
   }
 
-  document.addEventListener('keyup', onKeyUp, true);
+  /**
+   * Add a group of listeners to detect a fine-grained pointing device.
```
alice: touch isn't fine grained, is it? I think this is just "usage of any pointing device".

src/focus-ring.js
```diff
@@ -128,26 +131,75 @@ function init() {
...
+   * These listeners will be added when the polyfill first loads, and if
+   * the window is blurred and regains focus.
```
alice: s/and if the window is blurred and regains focus/and any time the window is blurred, so that they are active when the window regains focus./

src/focus-ring.js
```diff
@@ -128,26 +131,75 @@ function init() {
...
+    document.addEventListener('pointerdown', onInitialPointerMove);
+    document.addEventListener('pointerup', onInitialPointerMove);
+    document.addEventListener('touchmove', onInitialPointerMove);
+    document.addEventListener('touchstart', onInitialPointerMove);
+    document.addEventListener('touchend', onInitialPointerMove);
+  }
+
+  /**
+   * When the polfyill first loads, assume the user is in keyboard modality.
+   * If any event is received from a fine-grained pointing device (mouse, pointer, touch),
```
alice: Again, touch isn't fine-grained.

src/focus-ring.js
alice: https://drafts.csswg.org/mediaqueries-4/#valdef-media-pointer-coarse

"Examples include touchscreens ..."

src/focus-ring.js
```diff
@@ -128,26 +131,75 @@ function init() {
...
+   * turn off keyboard modality.
+   * This accounts for situations where focus enters the page from the URL bar.
+   * In that scenario, the keydown event is inconsistent, so we can't use it to detect modality.
```
alice: I would delete this and the following two lines - more justification than necessary.

test/index.js
```diff
@@ -85,7 +85,7 @@ function runMocha(driver, mocha) {
  */
 function browserFilter(browser) {
   return browser.getReleaseName() === 'stable'
-    && ['firefox', 'chrome'].includes(browser.getId());
+    && ['chrome', 'firefox'].includes(browser.getId());
```
alice: 🤔

test/fixtures/contenteditable-textbox.html
```diff
@@ -20,6 +20,12 @@
     </style>
   </head>
   <body>
+    <!--
+      FF won't focus a div with contenteditable if it's the first element on the page.
+      https://jsbin.com/cawevo/edit?html,output
+      Add a dummy button so Selenium can use it as the first focus target.
+    -->
+    <button id="start">Dummy</button>
```
alice: It confused me a little that the id was "start" but the contents was "Dummy".

test/specs/contenteditable-textbox.js
```diff
@@ -1,18 +1,26 @@
...
-    return fixture('contenteditable-true.html');
+    return fixture('contenteditable-textbox.html');
   });
 
-  // This test seems to fail because FF won't focus a div with contenteditable
-  // if it's the first element on the page. Weird.
-  // https://jsbin.com/cawevo/edit?html,output
-  it.skip('should apply .focus-ring on keyboard focus', function() {
-    return matchesKeyboard();
+  // FF won't focus a div with contenteditable if it's the first element on the page.
```
alice: That sounds like it was a lot of fun to debug 🙄

src/focus-ring.js
```diff
@@ -1,17 +1,12 @@
...
 
 /**
  * https://github.com/WICG/focus-ring
  */
 function init() {
+  var hadKeyboardEvent = true;
   var elWithFocusRing;
 
   var inputTypesWhitelist = {
-    'radio': true,
```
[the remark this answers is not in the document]
robdodson: I don't think so. That would cause focus-ring to be applied on mouse click because these get checked during the focus event. The reverted version added them because it was checking only on keyUp.

src/focus-ring.js
```diff
@@ -70,32 +65,43 @@ function init() {
...
-   *    keyboard navigation.
-   * 2. Unlike `focus`, keyup` will fire when the user navigates from the
-   *    browser chrome into the document. (For more, see issue #15)
+   * On `keydown`, set `hadKeyboardEvent`, add `focus-ring` class if the
+   * key was Tab/Shift-Tab or Arrow Keys.
    * @param {Event} e
    */
-  function onKeyUp(e) {
+  function onKeyDown(e) {
+    const allowedKeys = [9, 37, 38, 39, 40];
```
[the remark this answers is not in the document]
robdodson: There's a comment above that indicates it's tab and arrow keys but yeah.

src/focus-ring.js
robdodson: Fixed

src/focus-ring.js
robdodson: Fixed

src/focus-ring.js
robdodson: Fixed

src/focus-ring.js
robdodson: Fixed

src/focus-ring.js
robdodson: Fixed

src/focus-ring.js
robdodson: Fixed

src/focus-ring.js
robdodson: Fixed

test/fixtures/contenteditable-textbox.html
robdodson: Fixed

test/index.js
robdodson: was just double checking something. selenium-assistant stopped killing firefox 57 properly and I wasn't sure if the order in which the browsers opened/closed had anything to do with it. turns out it doesn't. the ordering doesn't matter and the issue isn't actually a show stopper. it just means if you test locally you need to close firefox when you're done. doesn't seem to bother travis though. I filed https://github.com/GoogleChromeLabs/selenium-assistant/issues/105

test/specs/contenteditable-textbox.js
robdodson: Not as much fun as https://github.com/WICG/focus-ring/pull/68/files#diff-6d1daa3464793cab0c4c5d8eeb17e778R178

PR #73, opened 2017-11-15
Remove dist

robdodson:
Depends on #72.

This removes/ignores the `dist` folder and adds a note to the README that tells devs to _only_ use a version of the polyfill that has been published to npm. This is to prevent noise on our PRs (which always require rebuilding `dist`) and because we don't keep a stable master branch so using the source directly _may_ lead to breakages.

Adds a [`prepublishOnly` hook](https://docs.npmjs.com/misc/scripts#deprecation-note) to the `package.json` to always run our tests and create a build before `npm publish` completes.

Updates the demo to use [unpkg](https://unpkg.com) so the hosted page always runs the latest polyfill on npm. Also includes polyfill.io link in the demo to polyfill classList in old IE.


README.md
```diff
@@ -19,8 +19,23 @@ in situations in which the `:focus-ring` pseudo-selector should match.
 
 `npm install --save wicg-focus-ring`
 
+_We recommend only using versions of the polyfill that have been published to npm, rather than
+cloning the repo and using the source directly. This helps ensure the version you're using is stable
+and thoroughly tested. If you do want to build from source, make sure you clone the latest tag._
```
alice: Suggestion: newline before "If"

robdodson: I'm going to add that after this change cuz I'm scared to touch this wonky rebase again :D

Issue #76, opened 2017-11-17
Change from 'const' to 'var'


jeroencornelissen:
https://github.com/WICG/focus-ring/blob/f70c271888db99b6427ee77571e8fcaff085d196/src/focus-ring.js#L82

Can this `const` not be a `var` like all the other variables?

robdodson:
oops good catch!

robdodson:
Fixed on master. I'll put out a new 3.0.1 release :)

PR #77, opened 2017-11-17
Begrudgingly add eslint back

robdodson:
@alice PTAL.

I really dislike ESLint because it mixes style guide enforcement with error checking and is super wonky to configure.

Having said that, as soon as I published v3.0 [someone pointed out](https://github.com/WICG/focus-ring/issues/76) that I had snuck an untranspiled `const` in there.

This PR adds ESLint back but scopes it to only lint the `src/focus-ring.js` file, and to only check that we're not using ES2015 features. It does not try to do style guide enforcement via ESLint (instead it defers to Prettier). Finally, the PR adds a precommit hook to run ESLint against any file changes to `src/focus-ring.js` and bail if it fails.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,11 @@
+module.exports = {
+  env: {
+    browser: true
+  },
+  // The only thing we want ESLint to do is yell at us if we accidentally use ES2015+.
+  // This is because we don't want/need to use Babel in this project.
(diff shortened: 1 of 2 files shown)


robdodson:
giphy

alice:
Everything about this PR is excellent.

PR #90, opened 2017-12-12
Rename :focus-ring to :focus-visible

alice:
@robdodson PTAL - will wait for https://github.com/w3c/csswg-drafts/pull/2104 to go in obviously.

Also, after this will need to do a major release and rename the repo and NPM module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,15 +2,15 @@
 [![Greenkeeper badge](https://badges.greenkeeper.io/WICG/focus-ring.svg)](https://greenkeeper.io/)
 
 Based on the proposed CSS
-[`:focus-ring`](https://drafts.csswg.org/selectors-4/#the-focusring-pseudo)
+[`:focus-visible`](https://drafts.csswg.org/selectors-4/#the-focus-visible-pseudo)
 pseudo-selector,
-this prototype adds a `focus-ring` class to the focused element,
-in situations in which the `:focus-ring` pseudo-selector should match.
+this prototype adds a `focus-visible` class to the focused element,
+in situations in which the `:focus-visible` pseudo-selector should match.
 
 # Details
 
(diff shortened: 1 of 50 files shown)


robdodson:
I think it might be useful for the README to have a note somewhere that this was originally called `:focus-ring` but renamed to `:focus-visible`. We've already done a fair amount of devrelling telling folks about :focus-ring so anything we can do to migrate people's thoughts will be useful.

robdodson:
oh hm... we're also going to need to npm deprecate wicg-focus-ring and publish a new project, wicg-focus-visible.

Also, these projects would need to update:
- [focus-ring-polyfill](https://www.npmjs.com/package/focus-ring-polyfill)
- [postcss-focus-ring](https://github.com/jonathantneal/postcss-focus-ring)

Possibly others...?

robdodson:
Sorry I see that you commented that we need to rename the repo and npm module. I don't think you can actually rename an npm dependency. You have to deprecate it which logs a warning when people install, then you have to publish a different one.

ryuran:
@robdodson you're right. Rename the npm package is not possible but we can deprecate it and register this new version with a new name.

joe-watkins:
FF team in the know on this? `moz-focusring` https://caniuse.com/#search=focus-ring

robdodson:
@joe-watkins yeah we've spoken with them about this.

robdodson:
oh sorry! you mean are they aware of the name change (I thought you were asking if they're aware that this WICG repo exists). That's actually a good question, I'll ping them to make sure they've seen it.

robdodson:
Also I'm going to see about renaming this polyfill at the end of this week, hopefully rolling it out Friday or even Saturday so we don't break anyone and we have some space to deal with all the npm stuff.

PR #112, opened 2018-01-26
add support for ShadowDOM WIP

codyloyd:
This PR adds support for using this polyfill in a project that uses shadowDOM.  It was specifically tested using Polymer components.

2 primary changes were required to make this work.  First was a bit of code (placed at the beginning of the file) that allows document level `focus` listeners to listen inside of shadowDOM.  I got that code [here](https://gist.github.com/samthor/7b307408e73784971ef0fcf4a8af6edd#file-shadowlisten-js-L38).  To make the `blur` listener work, I moved the creation of that listener inside the onFocus function.

The second change was to change the onFocus and onBlur functions to use either the composedPath function or document.activeElement as was described [here](https://github.com/WICG/focus-visible/issues/28#issuecomment-296925625)

The CSS required for getting it to work is a bit different than described in the original README.  Specifically, you need to add this rule to any component that contains a focusable element.
```
:focus:not(.focus-visible) {
  outline: 0;
}
```
In the context of shadowDOM, the body level `.js-focus-visible` isn't particularly meaningful.

### DIFF
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -1,113 +1,4 @@
 /**
- * Copyright 2017 Google Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * https://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-/**
- * source: https://gist.github.com/samthor/7b307408e73784971ef0fcf4a8af6edd#file-shadowlisten-js-L38
- */
-
-/**
- * @fileoverview Listener to provide Shadow DOM focus events, even inside shadow roots
- *
- * Normally, listening for focus with Shadow DOM will just return the top-most 'host' with a shadow
- * root. That may not be a problem! But if you want to find the 'real' focused element...
- *
- * To use this library, add this after the script is included-
- *   document.addEventListener('focus', shadowFocusHandler, true);
- *
- * You can then listen for `-shadow-focus` events, which will return the furthest focused element-
- *   document.addEventListener('-shadow-focus', function(ev) {
- *     console.info('got focused element', ev.detail, '..inside outer-most shadow root', ev.target);
- *   });
- */
-var shadowFocusHandler = (function() {
-  var eventName = '-shadow-focus';
-  var dispatch = function(target) {
-    var args = { composed: true, bubbles: true, detail: target };
-    var customEvent = new CustomEvent(eventName, args);
-    target.dispatchEvent(customEvent);
-  };
-
-  if (!window.WeakSet || !window.ShadowRoot) {
-    return function(event) {
-      dispatch(event.target);
-    };
-  }
-
-  var focusHandlerSet = new WeakSet();
-
-  /**
-   * @param {Node} target to work on
-   * @param {function(!FocusEvent)} callback to invoke on focus change
-   */
-  function _internal(target, callback) {
-    var currentFocus = target; // save real focus
-
-    // #1: get the nearest ShadowRoot
-    while (!(target instanceof ShadowRoot)) {
-      if (!target) {
-        return;
-      }
-      target = target.parentNode;
-    }
-
-    // #2: are we already handling it?
-    if (focusHandlerSet.has(target)) {
-      return;
-    }
-    focusHandlerSet.add(target);
-
-    // #3: setup focus/blur handlers
-    var hostEl = target.host;
-    var focusinHandler = function(ev) {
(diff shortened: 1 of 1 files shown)


LarsDenBakker:
Can this be looked into? Would be great to get this support merged!

robdodson:
Sure I can look into this today.

robdodson:
@codyloyd quick question, have you tested that this works in IE/Edge when web components are polyfilled? I don't think Edge has WeakSet and I just noticed it's used in the shadow listener code. I haven't read far enough to know if that would be a dealbreaker or not but this line https://gist.github.com/samthor/7b307408e73784971ef0fcf4a8af6edd#file-shadowlisten-js-L40 gave me pause.

LarsDenBakker:
Could just use a WeakMap with dummy values for IE.

codyloyd:
I have not tested in IE/Edge.

robdodson:
Just thinking out loud as I review this...

I'm a bit concerned about [this line](https://github.com/WICG/focus-visible/pull/112/files#diff-bf74a65b572b79d701701a8b9b1e01aeR59). If I'm understanding it correctly, it will do a recursive lookup of the parentNode until it either hits a ShadowRoot, or gets the parent of document (which is null). I think this would mean that if you had some form elements that were nested 1000 elements deep in the Light DOM, moving focus between each element would incur a 1000 element lookup. Still trying to figure out if I'm misreading it or if there's a way to shortcut that.

robdodson:
I think event.composedPath()[0] will always be the actual even target, regardless of shadow roots. Maybe we could just listen to that? Lemme play around with it...

robdodson:
FYI, this seems to report the proper target of focus and blur events, even if that target is inside a shadow root.

```
document.addEventListener('focus', function(event) {
      let innerTarget = event.composedPath()[0];
      console.log('focus innerTarget', innerTarget);
    }, true);

    document.addEventListener('blur', function(event) {
      let innerTarget = event.composedPath()[0];
      console.log('blur innerTarget', innerTarget);
    }, true);
```

robdodson:
@codyloyd do you think you might be able to look into just modifying the onFocus and onBlur handler of the original source to check if there's an event.composedPath and if so, use event.composedPath()[0] as the target for both functions? I wonder if that may do the trick.

codyloyd:
@robdodson yep 😄 I'll jump on that in the morning

codyloyd:
Ok, I've been playing with this again this morning.  I may be misunderstanding something, but it seems like the `focus` listener is not always getting called on elements in the shadowDOM without the polyfill code I added. So, the event.composedPath() _is_ reporting the correct element, but the callback isn't always getting run when you click or tab around the page.  

In my test app I have a `<button>`, a div with `tabindex="0"` and an `<input>`. Without the polyfill, (and _with_ the `event.composedPath` code) tabbing/clicking through the page works correctly for the `<button>` but not for the `<div tabindex="0">` or the `<input>`.   

For the div and input, the onFocus callback isn't getting run at all (tested by simply adding  `console.log(e)` to the function), though you _can_ tab into the input and type.

It  _does_ seem to work when you focus/blur the window itself.  i.e. when I tab onto the `<div tabindex="0">`, nothing happens, but if I blur/focus the window (by switching tabs or something) the onFocus listener _is_ run and the correct element has the `.focus-visible` class.

codyloyd:
new idea -- I currently have this working, but haven't done a ton of testing yet.

you can get the correct element from listening to a keyUp event, so I added an onKeyUp function and listener like so:
```
  function onKeyUp(e) {
    var allowedKeys = [
      9, // TAB
      37, // LEFT
      38, // UP
      39, // RIGHT
      40 // DOWN
    ];
    
    // Ignore keypresses if the user is holding down a modifier key.
    if (e.altKey || e.ctrlKey || e.metaKey) {
      return;
    }

    // Ignore keypresses which aren't related to keyboard navigation.
    if (allowedKeys.indexOf(e.keyCode) === -1) {
      return;
    }

    var target =
      typeof e.composedPath === 'function'
        ? e.composedPath()[0]
        : document.activeElement;

    if (hadKeyboardEvent || focusTriggersKeyboardModality(target)) {
      target.addEventListener('blur', onBlur, true);
      addFocusRingClass(target);
      hadKeyboardEvent = false;
    }
  }
```
this is just a combination of the filtering code from `onKeyDown` and the focusing logic from `onFocus`

codyloyd:
I went ahead and updated the PR, removing the polyfill and adding the KB/Click listeners that I described in my previous comment.  I have tested on Firefox and Chrome, but not IE/Edge.

robdodson:
Looking into this today. sorry for the lag :P

robdodson:
So I think what we'll need to do is more complicated. As you discovered, only 1 focus event gets fired when you enter an element. Apparently this is by design (though it seems weird that keyup would also fire). I'm going to email one of the devs about this. I think it _sucks_.

To work around this we'll need to listen for the host of a focus event and see if it has a shadow root. If it does, we have to attach onFocus/onBlur listeners to it to catch its internal focus/blur events (they still fire, they just don't bubble up anymore).

Because you can have shadow dom within shadow dom this will be fun :P

We'll also need to tear the whole thing down if the user randomly clicks somewhere else on the page.

PR #117, opened 2018-02-07
Add tab switching fix. Fixes #115

### LINKED ISSUE #115: Focus-ring is visible after switching browser tabs
If you switch browser tabs, with the mouse or keyboard. The JS will add a `focus-visible` class and `data-focus-visible` attribute.

To me and our users, this is unexpected behaviour. What do you think?

Replicate by clicking the `div` with `tabindex="0"` on the [demo page](https://wicg.github.io/focus-visible/demo/). Now, no focus-ring is visible. Switch browser tabs. And when you switch back, the focus-ring is visible.

Thx

robdodson:
This adds a short timeout where we retain the fact that an element just had focus-visible applied.
If the user switches tabs during this interval, we hold on to this information, so when they return we can reapply the focus-visible class to the same element. The browser will actually handle calling focus on the element when we switch back to the previous tab, we just need to provide a signal to remind it that the focus came from the keyboard.

The only issue I've found with this is if you switch tabs _very_ fast in Firefox, it doesn't fire a `visibilitychange` event, so we can't properly reapply the ring. But this is the best I think we can do.

I also removed the onWindowFocus/Blur handlers because they were trying to do this same job but were firing on every focus or blur event which I don't think was intended.

I've added tests for Chrome and Firefox and have manually verified in Safari, Edge, and IE 11.

@alice PTAL

### DIFF
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -3,6 +3,8 @@
  */
 function init() {
   var hadKeyboardEvent = true;
+  var hadFocusVisibleRecently = false;
+  var hadFocusVisibleRecentlyTimeout = null;
   var elWithFocusRing;
 
   var inputTypesWhitelist = {
@@ -52,7 +54,7 @@ function init() {
    * the author.
    * @param {Element} el
    */
-  function addFocusRingClass(el) {
+  function addFocusVisibleClass(el) {
     if (el.classList.contains('focus-visible')) {
       return;
     }
@@ -65,7 +67,7 @@ function init() {
    * originally added by the author.
    * @param {Element} el
    */
-  function removeFocusRingClass(el) {
+  function removeFocusVisibleClass(el) {
     if (!el.hasAttribute('data-focus-visible-added')) {
       return;
     }
@@ -108,12 +110,13 @@ function init() {
    * @param {Event} e
    */
   function onFocus(e) {
-    if (e.target == document) {
+    // Prevent IE from focusing the document or HTML element.
+    if (e.target == document || e.target.nodeName == 'HTML') {
       return;
     }
 
     if (hadKeyboardEvent || focusTriggersKeyboardModality(e.target)) {
-      addFocusRingClass(e.target);
+      addFocusVisibleClass(e.target);
       hadKeyboardEvent = false;
     }
   }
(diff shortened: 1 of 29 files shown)


src/focus-visible.js
```diff
@@ -123,64 +126,34 @@ function init() {
...
+      }
+      addInitialPointerMoveListeners();
     }
   }
 
   /**
-   * Add a group of listeners to detect usage of any pointing devices.
-   * These listeners will be added when the polyfill first loads, and anytime
-   * the window is blurred, so that they are active when the window regains focus.
+   * Add a group of listeners to detect usage of any pointing devices. These
```
alice: This somehow got reflowed. We should agree on a width so this doesn't keep happening.

src/focus-visible.js
```diff
@@ -194,11 +167,23 @@ function init() {
...
+    document.removeEventListener('pointerdown', onInitialPointerMove);
+    document.removeEventListener('pointerup', onInitialPointerMove);
+    document.removeEventListener('touchmove', onInitialPointerMove);
+    document.removeEventListener('touchstart', onInitialPointerMove);
+    document.removeEventListener('touchend', onInitialPointerMove);
+  }
+
   /**
    * When the polfyill first loads, assume the user is in keyboard modality.
-   * If any event is received from a pointing device (e.g mouse, pointer, touch), turn off
```
alice: Why remove this? It's not intended to be a complete list and may not be.

src/focus-visible.js
```diff
@@ -123,64 +126,34 @@ function init() {
...
-    addInitialPointerMoveListeners();
-
-    if (document.activeElement.classList.contains('focus-visible')) {
-      // Keep a reference to the element to which the focus-visible class is
-      // applied so the focus-visible class can be restored to it if the window
-      // regains focus after being blurred.
-      elWithFocusRing = document.activeElement;
+  function onVisibilityChange(e) {
+    if (document.visibilityState == 'hidden') {
+      if (hadFocusVisibleRecently) {
```
alice: Can you add a comment explaining this logic?

test/fixtures/change-tabs-sometimes-match.html
```diff
@@ -0,0 +1,45 @@
...
+        margin: 20px;
+      }
+      [tabindex] {
+        width: 100px;
+        height: 60px;
+        padding: 3px;
+        border: 1px solid #000;
+      }
+      .js-focus-visible :focus:not(.focus-visible) {
+        outline: 0;
```
alice: Setting outline to 0 doesn't null the outline-color:
```js
$0.style.outline = 0
window.getComputedStyle($0).outlineColor
// "rgb(51, 51, 51)"
```

src/focus-visible.js
robdodson: cuz it was going to overflow :P i'll add it back

src/focus-visible.js
robdodson: oh i think i used to have .vscode project settings which may have had it at 100 but those weren't checked into github so I reformated to a default 80

test/specs/change-tabs-sometimes-match.js
```diff
@@ -0,0 +1,41 @@
...
+    return fixture('change-tabs-sometimes-match.html');
+  });
+
+  it('should retain .focus-visible if the user switches tabs and an element had .focus-visible from keyboard', async function() {
+    let body = await driver.findElement(By.css('body'));
+    let el = await driver.findElement(By.css('#el'));
+    await body.click();
+    await body.sendKeys(Key.TAB);
+    await el.sendKeys(Key.SPACE);
+    await driver.sleep(4000); // sleep while we open and close a new tab.
```
alice: Any chance we could wait for the visibility events to fire instead of this terrifying timeout?

test/fixtures/change-tabs-always-match.html
```diff
@@ -0,0 +1,39 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>change tabs always match fixture</title>
```
alice: Could be "change tabs with focus on element which always matches focus ring"?

test/fixtures/change-tabs-sometimes-match.html
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>change tabs sometimes match fixture</title>
```
alice: change tabs with focus on element which matches focus ring only on keyboard focus?

test/specs/change-tabs-sometimes-match.js
robdodson: oh this may actually be possible, I didn't know there was an executeAsyncScript command.

robdodson: http://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_WebDriver.html

robdodson: I tried mucking around with executeAsyncScript but it seems like it was blocking the next step in the driver instructions that actually opens the new window. I'm not in love with the `sleep()` either, Selenium is kinda awful. I'd like to just move on. I've run the test 3 times through travis and it works. I won't be surprised if it becomes flakey someday. At which point I made just change it to a `.skip()` and only run it locally as a sanity check or invest more time in debugging Selenium.

robdodson:
Alice LGTM'd over chat.

src/focus-visible.js
```diff
@@ -123,64 +126,44 @@ function init() {
...
-   * When the window regains focus, restore the focus-visible class to the element
-   * to which it was previously applied.
-   */
-  function onWindowFocus() {
-    // When removing the activeElement from DOM it's possible IE11 is in state
-    // document.activeElement === null
-    if (!document.activeElement) {
-      return;
+    if (e.target.classList.contains('focus-visible')) {
+      hadFocusVisibleRecently = true;
```
alice: Suggested comment here:

To detect a tab switch, we look for a blur followed rapidly by a visibility change. If we don't see a visibility change within 100ms, it's probably a regular focus change.

src/focus-visible.js
```diff
@@ -123,64 +126,44 @@ function init() {
...
-      // applied so the focus-visible class can be restored to it if the window
-      // regains focus after being blurred.
-      elWithFocusRing = document.activeElement;
+  function onVisibilityChange(e) {
+    if (document.visibilityState == 'hidden') {
+      // If the tab becomes active again, the browser will handle calling focus
+      // on the element (Safari actually calls it twice).
+      // At this point we just need to remind the system whether or not the user
+      // previously used a keyboard to focus the element.
+      if (hadFocusVisibleRecently) {
```
alice: Suggested comment here:

If this tab change caused a blur on an element with `focus-visible`, re-apply the class when the user switches back to the tab.

robdodson:
@alice i've added both of the above comments in master

PR #118, opened 2018-02-07
Any focus proceeded by a keypress should trigger focus-visible.

robdodson:
@alice PTAL


test/fixtures/programmatic-focus.html
```diff
@@ -0,0 +1,26 @@
...
+      }
+    </style>
+  </head>
+  <body>
+    <button id="start">Focus other button</button>
+    <button id="el">Focus me!</button>
+    <script src="/dist/focus-visible.js"></script>
+    <script>
+      start.onclick = function() {
+        el.focus();
```
alice: Can you also add a test for focusing in a new microtask?

robdodson: sure

robdodson:
Ignore the conflicts. I think I just forgot to rebase against latest master where the most recent change was some additional comments.

src/focus-visible.js
```diff
@@ -76,37 +76,39 @@ function init() {
...
+   * This is useful for determining if clicking on an element calls focus() on a
+   * different element.
+   * @param {Event} e
+   */
+  function onPointerDown(e) {
+    forceNoFocusVisible = false;
+    elementToBePointerFocused = e.target;
+    // Clicking inside of a <select multiple> element will fire a mousedown
+    // on the <option> child, but focus will be set on the <select> itself.
+    if (elementToBePointerFocused.nodeName == 'OPTION') {
```
robdodson: Note. If we keep this change then we'll need to do something different for elements inside of shadow dom.

robdodson: I think what we can do there is .getRootNode() or whatever the API is called. If that returns a shadowRoot, we'll need to set this to the element's host.

robdodson: https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode

robdodson:
I went back to the previous approach. I moved the pointer exclusion stuff to [a separate branch](https://github.com/WICG/focus-visible/tree/exclude-pointers) in case we ever want to use it.

I also added a fix and test for a small bug I found. If you press a key on an already focused element, the polyfill thinks it's in keyboard modality but it never gets cleared out because there's never a focus event. If you then click on something with a mouse, we still think we're in keyboard modality. I just added a handler for any mouse/pointer/touch down event that forces keyboard modality off.

Ready for review when you have time.

karlhorky:
First of all, thanks for this library. Really liking the polyfill efforts from standards orgs.

What issue does this pull request solve? It seems to be the cause of unusual behavior when using <kbd>cmd</kbd>-<kbd>tab</kbd> to switch away from browser windows:

1. User has used the mouse to select elements. No focus styles applied.
2. User hits the <kbd>cmd</kbd> key for a fraction of a second before hitting <kbd>tab</kbd> to switch to another window.
3. Focus styles appear for this fraction of a second, causing a confusing flash of styles before the window switches.

---

A simple reproduction can be seen on [the demo page](https://wicg.github.io/focus-visible/demo/), where a press on the <kbd>cmd</kbd> key after mouse-only selection causes the focus style to be applied:

Kapture 2019-04-24 at 11 56 07

The flashing of styles while using <kbd>cmd</kbd>-<kbd>tab</kbd> does not appear in version `4.0.1` of the library.

---

If there's a very important reason that the changes from this pull request were made, maybe there can be a smaller blacklist of modifier keys (window and tab switching keys, for example) that it won't trigger on?

karlhorky:
I can also open this as a new issue, if that would be more appropriate.

robdodson:
Hey @karlhorky,

I plan to follow up more but real quick, i think you'd be interested in this:
https://github.com/WICG/focus-visible/pull/184

We updated the heuristic so meta keys don't trigger keyboard focus. Specifically to avoid when folks are using keyboard shortcuts.

karlhorky:
Ah cool, you're on it already! Nice 🙌. Watching now.

PR #123, opened 2018-02-16
Fix the link to focus-visible.js

vkrol:


robdodson:
actually would you mind changing this so it just points to http://wicg.github.io/focus-visible ?  We dont' want folks downloading the raw source.

vkrol:
I thought that the main idea was to show people how “tiny” the polyfill is.

robdodson:
@vkrol Maybe that was the original intent, but we've seen folks directly downloading the source which we want to avoid because it may be in an unstable state. We only want folks to download tagged versions that have been published to npm.

PR #129, opened 2018-03-12
Add minified bundle

boriskor:
It would be nice for this polyfill to have a minified bundle, so I've added it 🙂


robdodson:
rad thank you!

boriskor:
Thanks for a quick response! When do you plan to make the next publish? 😇

robdodson:
It should be live now

boriskor:
Thank you, sir!

PR #133, opened 2018-03-30
Add Sauce tests for Edge and IE11.

robdodson:


robdodson:
It creates a pretty big modal for these inputs, here's the one for date:

screen shot 2018-03-29 at 6 19 28 pm

And it seems to draw its own focus indicator type thing inside of the modal.

PR #139, opened 2018-04-25
Add check for availability of `document` browser variable

nelonoel:
Assuming the `document` global variable exists can cause build errors on static pre-renderers. Added a simple `typeof` check before initialization.

Here's a related document on GatsbyJS' website:
https://www.gatsbyjs.org/docs/debugging-html-builds


robdodson:
Thank you!

PR #140, opened 2018-05-13
Check if element has `data-focus-visible-added` attribute on blur

darrylblake:
This is to solve an issue with React, if the classes are updated on render the `focus-visible` class is removed as React is unaware of it. This could happen when the element should still be focused, one example being a switch component that's toggled with the keyboard updating the classes.

### DIFF
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -146,7 +146,10 @@ function init() {
       return;
     }
 
-    if (e.target.classList.contains('focus-visible')) {
+    if (
+      e.target.classList.contains('focus-visible') ||
+      e.target.hasAttribute('data-focus-visible-added')
+    ) {
       // To detect a tab/window switch, we look for a blur event followed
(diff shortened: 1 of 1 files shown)


src/focus-visible.js
```diff
@@ -146,7 +146,10 @@ function init() {
       return;
     }
 
-    if (e.target.classList.contains('focus-visible')) {
+    if (
+      e.target.classList.contains('focus-visible') ||
+      el.hasAttribute('data-focus-visible-added')
```
robdodson: It doesn't look like `el` is in scope here. Did you mean `e.target`?

darrylblake: I did indeed! Sorry copypasta... fixed.

robdodson:
looks good to me. I can merge once travis gets back to us :)

darrylblake:
Great thanks! :)

PR #143, opened 2018-05-15
Making it works when loaded async or defer

dimitrinicolas:
Testing if `document.readyState` is equal to `'compete'` or `'interactive'` (and not only `'complete'`). So the script can be loaded with an async and/or defer script tag.


robdodson:
Great, thank you :)

PR #146, opened 2018-05-17
Fixes #80. Guard against IE11 focusing SVG.

### LINKED ISSUE #80: TODO: Add test to catch IE null activeElement regressions
See #56 for an example. It'd be good if we could have a test suite check for this.

robdodson:


robdodson:
@alice does this look ok to you?

test/specs/svg.js
```diff
@@ -0,0 +1,44 @@
...
+describe('svg focus', function() {
+  beforeEach(function() {
+    return fixture('svg.html');
+  });
+
+  it('should NOT apply .focus-visible if a non-interactive SVG is keyboard focused', async function() {
+    async function tabUntil(body, end) {
+      let el;
+      while (true) {
+        focused = await driver.executeScript(`
```
alice: Why a template literal here?

src/focus-visible.js
```diff
@@ -25,15 +25,17 @@ function init() {
...
    * @param {Element} el
    */
   function isValidFocusTarget(el) {
     if (
       el &&
       el !== document &&
       el.nodeName !== 'HTML' &&
-      el.nodeName !== 'BODY'
+      el.nodeName !== 'BODY' &&
+      el.classList &&
```
alice: can we do `'classList' in el && 'contains' in el.classList` to avoid breaking my brain?

test/specs/svg.js
```diff
@@ -0,0 +1,44 @@
...
+          return document.activeElement.id
+        `);
+        if (focused === end) {
+          break;
+        }
+        if (focused) {
+          // IE11's selenium driver won't move focus if we send it to body again
```
alice: This whole block looks like the result of a lot of crying and furious Googling.

test/specs/svg.js
```diff
@@ -0,0 +1,44 @@
...
+          await body.sendKeys(Key.TAB);
+        }
+      }
+    }
+
+    let body = await driver.findElement(By.css('body'));
+    await body.click();
+    await tabUntil(body, 'end');
+    let actual = await driver.executeScript(`
+      return window.getComputedStyle(document.querySelector('#end')).outlineColor
```
alice: Can you add a comment explaining how this tests for the failure mode you describe at the top of the file?

test/specs/svg.js
robdodson: All of the tests that manually execute script through webdriver use a template literal. https://github.com/WICG/focus-visible/blob/master/test/specs/programmatic-focus-always.js#L32-L34

robdodson: Blargh, GitHub submitted my reply before I was finished typing.

I originally used template literals because the code was copied from another project and we had done it there. I can update all of the tests if you'd prefer that.

src/focus-visible.js
robdodson: sure

test/specs/svg.js
robdodson: It was mostly just swearing and trying stuff. Googling was no help.

robdodson:
made some changes...let's see if sauce still passes 🤞

robdodson:
womp womp

jasononeil:
Thanks for this! We were getting a flood of `Unable to get property 'contains' of undefined or null reference` errors in IE11 and this was the root cause. While we wait for this to be merged we've used `focusable=false` on our inline SVG tags.

Is there anything I can help with to get this merged? It looks like the tests timed out on Edge

robdodson:
oh I totally forgot this had not landed yet! I think the only thing holding it back is my poorly written test. I may break this PR in half because I'm confident the fix works, but getting the test to run reliable in SauceLabs is weird. Let me try that, we can land the fix at least, and I'll do a separate PR for the test.

robdodson:
oh sweet, it looks like the tests actually do pass. The previous failure was just Sauce Labs timing out. I misread the output :P

Merging!

jasononeil:
Great, thanks so much!

PR #150, opened 2018-06-15
use proper property name when checking for read-only elements.

plinehan:
introduced in 66a7ce1fff178662badd02c2fece0c38ac88a3a3 when converting from selectors.


robdodson:
Good catch. It seems like we should add a test for this as well.

PR #153, opened 2018-06-19
Fixes #152. Show :focus-visible if screen reader is running.

### LINKED ISSUE #152: `focus-visible` class is not applied when using a screen reader.
When navigating with a screen reader, e.g. when pressing CMD-OPT-RIGHT in VoiceOver, the `focus-visible` class is not applied.  The problem is that `onFocus` sets `hadKeyboardEvent = false` so the subsequent focus (with no intervening keydown event, because the screen reader swallowed it) doesn't realize the user is still using a keyboard.

This is different from the default behavior on Chrome.

robdodson:
This might need more testing but it seemed to pass all of our local and sauce tests.

### DIFF
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -132,7 +132,6 @@ function init() {
 
     if (hadKeyboardEvent || focusTriggersKeyboardModality(e.target)) {
       addFocusVisibleClass(e.target);
-      hadKeyboardEvent = false;
     }
   }
 


robdodson:
Assigning to @alice to take a look when she gets back from vacation.

src/focus-visible.js
```diff
@@ -132,7 +132,6 @@ function init() {
 
     if (hadKeyboardEvent || focusTriggersKeyboardModality(e.target)) {
       addFocusVisibleClass(e.target);
-      hadKeyboardEvent = false;
```
alice: Why this change?

robdodson: In my testing, it seemed like the screen reader swallowed keydown events. Instead, focus would just move to an element, which would then hit the line above and set `hadKeyboardEvent` to false. Meaning the next time the screen reader moved focus, it would seem as if it came from a mouse.

robdodson: Blargh. I hate that GitHub now submits comments when I hit shift-enter.

Continuing from the above...

The polyfill currently starts with `hadKeyboardEvent` set to `true`. And if we ever detect any mouse event, we set it to `false` (https://github.com/WICG/focus-visible/blob/master/src/focus-visible.js#L116).

So removing this line makes the keydown-swalling-screen-reader scenario work. And it doesn't seem to break anything because `onPointerDown` should handle turning `hadKeyboardEvent` off if the user ever switches to mouse.

alice: Does this result in being permanently stuck in focus-visible mode once any key is pressed?

alice: Apologies for the crossed comment trajectories.

Makes sense - this would probably also mean that a keypress that causes a modal to open would also stay in focus-visible mode.

PR #155, opened 2018-06-21
chore: use npm ci

DanielRuf:


.travis.yml
```diff
@@ -7,8 +7,11 @@ node_js:
   - "8"
 cache:
   directories:
-    - node_modules
+    - "$HOME/.npm"
```
robdodson: Can you help me understand this change? I'm not super familiar with the npm cache and reading their docs didn't help much.

DanielRuf: `npm ci` always deletes `node_modules` so we can just use the npm cache ;-)

robdodson:
cool thanks!

PR #157, opened 2018-06-21
chore: add Node.js 10

DanielRuf:


robdodson:
hm it looks like the lock file is out of sync according to travis. I'll see if I can fix this.

DanielRuf:
> hm it looks like the lock file is out of sync according to travis. I'll see if I can fix this.

Should be fine in master atm after my last merged commit.

robdodson:
ok let's give it a shot :)

PR #160, opened 2018-06-28
Use `isContentEditable` rather than checking the value of `contentEdi…

plinehan:
…table`.

Handles values like `plaintext-only` properly.  Verified that Chrome also shows focus for `plaintext-only`.  Did not check other values.

### DIFF
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -58,7 +58,7 @@ function init() {
       return true;
     }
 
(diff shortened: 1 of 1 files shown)


robdodson:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

robdodson:
Thanks!

PR #164, opened 2018-08-01
Don't apply if element is readonly or disabled.

robdodson:
I noticed that when you tab to a `[readonly]` element it was still applying `:focus-visible`. That's because [the `onfocus` handler](https://github.com/WICG/focus-visible/blob/f0a0f2d8dc0619debc4ab24711871dd25d05454f/src/focus-visible.js#L135) will short circuit the readonly checks on lines 53 and 57 if focus was preceded by a keydown event.

To work around this I've moved `readOnly` and `disabled` checks to the isValidFoucsTarget handler that proceeds everything. @alice wdyt?

### DIFF
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -32,7 +32,9 @@ function init() {
       el &&
       el !== document &&
       el.nodeName !== 'HTML' &&
-      el.nodeName !== 'BODY'
+      el.nodeName !== 'BODY' &&
+      !el.readOnly &&
+      !el.disabled
     ) {
       return true;
     }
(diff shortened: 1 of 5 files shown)


robdodson:
Blargh... looks like tests are failing in Firefox because if you call sendKeys in Selenium and the element is disabled it throws an exception. [shakes fist at Selenium]

alice:
I think it's weird that readonly fields don't get a focus ring by default, honestly.

Disabled fields can't be focused even with `focus()`, so I'm fine making them not a valid focus target.

robdodson:
Filed a Chromium bug for the `readOnly` thing.
https://bugs.chromium.org/p/chromium/issues/detail?id=870062

robdodson:
Gonna close this and instead see if we can get Chrome's behavior to be less weird re: readOnly focus ring drawing. See the above crbug for more details.

PR #168, opened 2018-08-07
See if we can use greenkeeper-lockfile

robdodson:
Our greenkeeper builds keep failing because when greenkeeper proposes a change to package.json, it needs to also update the lockfile. Otherwise `npm ci` fails.


robdodson:
Closed in favor of #169

Issue #163, opened 2018-07-24
How to import the init function from focus-visible? Can it be a module?


renojvarghese:
Hi. I was wondering how I could import just the init function.  I am using this polyfill as a node module, and I have some things that need to be checked before utilizing the focus-visible polyfill. I was looking for a module.exports, but could not find one in the src code

robdodson:
We did not structure the original polyfill as a module, but with the widespread support of ES Modules we should probably update it to be one.

A temporary workaround if you don't feel like modifying the source would be to include the script async using a bit of JavaScript.

```js
var polyfill = document.createElement('script'); 
polyfill.src = "https://my-site.com/scripts/focus-visible.min.js";
var script = document.getElementsByTagName('script')[0];
script.parentNode.insertBefore(polyfill, script);
```

renojvarghese:
Thanks. Works well, but I ended copying and pasting the init function into my own js file. Really enjoyed this polyfill. It was exactly what I was looking for in that I didn't have to decide between one user experience over another

robdodson:
Glad to hear it helped! I'll leave this issue open and edit the title so we know to come back and work on an exportable module.

renojvarghese:
Mind if I take a stab at it? I think I might know how to do it such that it will work in both the browser and as a node module

robdodson:
@renojvarghese sure thing! Maybe you were thinking the same thing, but we'd want it to be an ES Module so folks could use standard `import` to pull it in.

renojvarghese:
@robdodson ^^^ Is this what you mean?

robdodson:
Hi @renojvarghese, sorry for my total lack of oversight, but I just realized that we're already exporting the distribution of focus-visible with a module friendly wrapper (look at the source for dist/focus-visible.js).

Also, to better answer your original question... rather than importing the init function, I would recommend dynamically importing the entire module. You can do this with ES modules using dynamic [`import()` syntax](https://developers.google.com/web/updates/2017/11/dynamic-import)

```
if (someCondition) {
  import('./node_modules/focus-visible/dist/focus-visible.min.js')
	.then(_ => console.log('focus-visible loaded!'));
}
```

If you're using Node's `require` syntax, it has a similar dynamic mechanism.

```
if (someCondition) {
  require('focus-visible');
}
```

[...]

PR #174, opened 2018-08-08
making focus-visible an es module

renojvarghese:
made focus-visible a module using https://www.matteoagosti.com/blog/2013/02/24/writing-javascript-modules-for-both-browser-and-node/ as a guide

for https://github.com/WICG/focus-visible/issues/163


robdodson:
Thanks for working on this! Please see my comment in #163. Sorry for any confusion!

Issue #167, opened 2018-08-07
Update explainer heuristic. Consider ignoring meta, alt/option, control keys.


robdodson:
## Current behavior

This is currently what's in the explainer [example heuristic](https://github.com/WICG/focus-visible/blob/master/explainer.md#example-heuristic)

> The heuristic used to decide the current modality should not be defined normatively. An example heuristic is to update modality on each style recalc: if the most recent user interaction was via the keyboard; and the key pressed was either Tab, Shift + Tab, or an arrow key; then the modality is keyboard. Otherwise, the modality is not keyboard.

The above snippet does not match the polyfill or the experimental implementation in Chrome—both of which turn on keyboard modality if focus was proceeded by _any_ key press (including meta and modifier keys).

## Proposed behavior

@alice and I were discussing if this behavior should be changed so any key would trigger keyboard modality _unless_ the keypress included a meta key (⌘ Command / ⊞ Windows), option / alt key, or control key.

Modifier keys, like shift, would still trigger keyboard modality (for instance, shift + Tab).

The example heuristic text would be updated to something like:

> The heuristic used to decide the current modality should not be defined normatively. An example heuristic is to update modality on each style recalc: if the most recent user interaction was via the keyboard; **and the key press did not include a meta, alt/option, or control key**; then the modality is keyboard. Otherwise, the modality is not keyboard.

## TODOs

- [ ] Update the explainer's example heuristic.
- [ ] Update Chrome's `:focus-visible` impl. to match.
- [ ] Update the polyfill to match.

@alice, please let me know if I'm mistaken about any of the above 😸

alice:
I think that example heuristic text looks good :+1: 

I can work on updating the Chrome native implementation and adding some WPTs.

robdodson:
Assigning to myself to create CRBug for chrome impl.

bradkemper:
The option key on Mac doesn’t seem to be triggering keyboard mode. 

In Safari, there is a checkbox setting to “press tab to highlight each item on a web page (including links and all form items, instead of just text fields and pop ups (drop downs)). Using the option key reverts that behavior temporarily. If the checkbox is not checked, using the option key will temporarily make it act as though it is checked. 

Chrome on Mac has a similar setting. 

But using the option key in this way isn’t showing the focus ring the way it does without it.

alice:
> The option key on Mac doesn’t seem to be triggering keyboard mode.

[...]

PR #184, opened 2018-09-22
Meta keys should not match.

robdodson:
Addresses heuristic update in #167.

@alice PTAL when you have a chance 😸

### DIFF
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -98,12 +98,18 @@ function applyFocusVisiblePolyfill(scope) {
   }
 
   /**
-   * Treat `keydown` as a signal that the user is in keyboard modality.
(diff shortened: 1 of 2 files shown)


alice:
:+1: once we figure out what's up with the tests.

robdodson:
@cdata 	I'd like to land this before releasing v5 since it's also a breaking change. I'll add it to my calendar to get this sorted tomorrow.

karlhorky:
Is there any timeline on when v5 will be released?

Would like to upgrade to this whenever that drops, so that I don't need to [stay pinned to `4.0.1`](https://github.com/WICG/focus-visible/pull/118#issuecomment-486157073).

robdodson:
I can try to roll it out tomorrow

PR #194, opened 2019-04-25
chore(package): Fix false positive fail in test

eps1lon:
Test were exiting with non-zero for me locally. Probably due to test:server requiring a kill which exited with SIGTERM. Since we're only interested in the exit code of test:selenium we can use https://github.com/kimmobrunfeldt/concurrently/issues/127#issuecomment-364685336


robdodson:
Sounds good, thanks!

PR #196, opened 2019-05-01
Shadow DOM support

### LINKED ISSUE #28: Shadow DOM
#### Description
The Focus-ring Polyfill has some issues when used with shadow DOM. The Polyfill is not able to add the focus-ring class to focusable elements within the shadow DOM such as a buttons. This is due to the shadow DOM event model. Events bubble up from shadow DOM but the events are re-targeted. This means that the events look like they come from the host instead of its internal events within the shadow DOM.

#### Versions & Environment
Focus-ring: v1.0.1

#### Steps to Reproduce
- Clone repository `git@github.com:emilbillberg/focus-ring-shadow-dom.git`
- Start a local server such as live-server or similar
- The Polyfill is NOT able to set the focus-ring class on the button element
- Go to the index.html file and change the Polymer Global setting to `shady` instead of `shadow`
- The Polyfill is able to set the focus-ring class on the button element

#### Expected Results
The Polyfill should be able to set the focus-ring class on the button element when using Shadow DOM

#### Actual Results
The Polyfill sets the focus-ring class on the top custom element. In the example above this will be the custom element `my-app`.

### LINKED ISSUE #193: Is DOM ready necessary?
The only line that seems to depend on some sort of DOM readiness is:

```js
document.body.classList.add('js-focus-visible');
```

Could that be:

```js
document.documentElement.classList.add('js-focus-visible');
```

…so this script could be loaded lazily or async?

cdata:
# Shadow DOM support

This change enables custom elements to coordinate with the polyfill in order to apply it to discrete scopes of the composed tree (such as their shadow roots). Highlights include:

 - The polyfill is now exported to `window` as a function that can be invoked on any `Document|ShadowRoot`
 - The polyfill is automatically, synchronously applied to the top-level `document` when its script runs, preserving the existing behavior
 - Code that wishes to use the polyfill in a `ShadowRoot` can apply it as needed
   - Shadow hosts that have the polyfill applied are decorated with a `data-js-focus-visible` attribute
 - A `focus-visible-polyfill-ready` event is now dispatched on `window` to support lazy loading scenarios (see discussion in #193)

## Example usage as a mixin

In the course of implementing this proposal, I also crafted a mixin to demonstrate how a custom element should coordinate with the polyfill in order to take advantage of `:focus-visible`-like capabilities.

I am **not** proposing to include the mixin in the main project, but it is linked here for future reference:

https://github.com/cdata/focus-visible/blob/fc33395294dd69ad8af70e822e6176a8893816e2/src/custom-element-mixin.js#L1-L49

## Compared to #112 

There is an older WIP proposal (#112) up for consideration that attempts to add transparent support for arbitrary shadow roots. This approach would ultimately require that `focus` and `blur` be observed at every shadow boundary throughout the composed tree in order to work transparently.

The design in the proposal before you expects a Shadow DOM-using component author to opt-in to the polyfill. This allows the polyfill to be applied to Shadow DOM-using components that need it, while avoiding the cost of observing changes in shadow roots that do not know about or do not want to use the polyfill.

## Examples

For the purposes of the following examples:

 - `:focus` elements are styled with purple `color` and `border-color`
 - `:focus-visible` elements receive the default focus ring
 - Elements with shadow roots have `delegatesFocus: true` configured
 - Custom elements apply the [mixin](https://github.com/cdata/focus-visible/blob/fc33395294dd69ad8af70e822e6176a8893816e2/src/custom-element-mixin.js#L1-L49) referenced above to coordinate with the polyfill

This change w/ pointer traversal | This change w/ keyboard traversal
-------------------------------- | ----------------------------------
focus-not-visible | focus-visible

### Shadow root styling problem

Without this change, it is not possible to use the polyfill from within a shadow root. In the following example, you can see that the top-level document elements can be styled to hide the focus ring when focusing with pointer events, but focusable elements inside of a shadow root still display a focus ring:

Failure to style shadow root-encapsulated elements |
---------------------------------------------------- |
focus-not-visible-broken |

Fixes #193 
Fixes #28 
Fixes #112

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,8 @@ this polyfill imperatively to the component's shadow root:
 
 ```javascript
 // Check for the polyfill:
-if (self.applyFocusVisiblePolyfill != null) {
-  self.applyFocusVisiblePolyfill(myComponent.shadowRoot);
+if (window.applyFocusVisiblePolyfill != null) {
+  window.applyFocusVisiblePolyfill(myComponent.shadowRoot);
 }
 ```
 
@@ -120,7 +120,7 @@ In order to act at the right time, you can observe the global
 
 ```javascript
 window.addEventListener('focus-visible-polyfill-ready',
-    () => self.applyFocusVisiblePolyfill(myComponent.shadowRoot),
+    () => window.applyFocusVisiblePolyfill(myComponent.shadowRoot),
     { once:  true });
 ```
 
--- a/README.md
+++ b/README.md
@@ -95,6 +95,42 @@ You can use a service like [Polyfill.io](https://polyfill.io) to download only t
 <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=Element.prototype.classList"></script>
 ```
 
+### Shadow DOM
+
+It could be very expensive to apply this polyfill automatically to every shadow
+root that is created in a given document, so the polyfill ignores shadow roots
+by default. If you are using Shadow DOM in a component, it is possible to apply
+this polyfill imperatively to the component's shadow root:
+
+```javascript
+// Check for the polyfill:
+if (self.applyFocusVisiblePolyfill != null) {
+  self.applyFocusVisiblePolyfill(myComponent.shadowRoot);
+}
+```
+
+### Lazy-loading
+
+When this polyfill is lazy-loaded, and you are applying the polyfill to a shadow
+root with JavaScript, it is important to know when the polyfill has become
+available before trying to use it.
+
+In order to act at the right time, you can observe the global
+`focus-visible-polyfill-ready` event:
+
+```javascript
+window.addEventListener('focus-visible-polyfill-ready',
+    () => self.applyFocusVisiblePolyfill(myComponent.shadowRoot),
+    { once:  true });
+```
+
+**Important:** this event is _only_ intended to support late application of the
+polyfill in lazy-loading use cases. Do not write code that depends on the event
+firing, as it is timing dependent and only fired once. If you plan to lazy-load
+the polyfill, it is recommended that you check for it synchronously (see example
+above under "Shadow DOM") and listen for the event only if the polyfill isn't
+available yet.
+
 # Backwards compatibility
 Until all browsers ship `:focus-visible` developers will need to use it defensively to avoid accidentally
 removing focus styles in legacy browsers. This is easy to do with the polyfill.
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -252,17 +252,18 @@ function applyFocusVisiblePolyfill(scope) {
   scope.addEventListener('focus', onFocus, true);
   scope.addEventListener('blur', onBlur, true);
 
-  if (scope.nodeType === Node.DOCUMENT_NODE) {
-    document.documentElement.classList.add('js-focus-visible');
-
-    // We detect that a node is a ShadowRoot by ensuring that it is a
-    // DocumentFragment and also has a host property. This check covers native
-    // implementation and polyfill implementation transparently.
-  } else if (scope.nodeType === Node.DOCUMENT_FRAGMENT_NODE && scope.host) {
+  // We detect that a node is a ShadowRoot by ensuring that it is a
+  // DocumentFragment and also has a host property. This check covers native
+  // implementation and polyfill implementation transparently. If we only cared
+  // about the native implementation, we could just check if the scope was
+  // an instance of a ShadowRoot.
+  if (scope.nodeType === Node.DOCUMENT_FRAGMENT_NODE && scope.host) {
     // Since a ShadowRoot is a special kind of DocumentFragment, it does not
     // have a root element to add a class to. So, we add this attribute to the
     // host element instead:
     scope.host.setAttribute('data-js-focus-visible', '');
+  } else if (scope.nodeType === Node.DOCUMENT_NODE) {
+    document.documentElement.classList.add('js-focus-visible');
   }
 }
 
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -1,7 +1,11 @@
 /**
- * https://github.com/WICG/focus-visible
+ * Applies the :focus-visible polyfill at the given scope.
+ * A scope in this case is either the top-level Document or a Shadow Root.
+ *
+ * @param {(Document|ShadowRoot)} scope
+ * @see https://github.com/WICG/focus-visible
  */
-function init() {
+function applyFocusVisiblePolyfill(scope) {
   var hadKeyboardEvent = true;
   var hadFocusVisibleRecently = false;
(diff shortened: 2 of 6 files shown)


src/focus-visible.js
```diff
@@ -230,45 +234,57 @@ function init() {
...
+  if (scope.nodeType === Node.DOCUMENT_NODE) {
+    document.documentElement.classList.add('js-focus-visible');
 
-  if (['interactive', 'complete'].indexOf(document.readyState) >= 0) {
-    callback();
-  } else {
-    loaded = false;
-    document.addEventListener('DOMContentLoaded', load, false);
-    window.addEventListener('load', load, false);
+    // We detect that a node is a ShadowRoot by ensuring that it is a
```
alice: Putting this comment inside the previous if block is breaking my brain a little bit, though I can see why you did. Maybe switch the order of these if/else statements? It would be _marginally_ less efficient for the "normal" case, I guess, but more readable.

Or, perhaps a switch statement over `scope.nodeType` ?

test/specs/shadow-dom.js
```diff
@@ -0,0 +1,65 @@
...
+
+  it('should apply .focus-visible on keyboard focus', async function() {
+    return matchesKeyboard();
+  });
+
+  it('should NOT apply .focus-visible on mouse focus', function() {
+    return matchesMouse(false);
+  });
+
+  describe('focusable elements in shadow', () => {
```
alice: Seems like this test is failing in CI for some reason.

alice:
Generally looking really good (that gif blew my mind), not sure what's going on with that test though.

src/focus-visible.js
cdata: Thanks @alice I'll try to rework this so that it is less of an eyesore

test/specs/shadow-dom.js
cdata: Sorry about that, will take a look

alice:
Just had a quick look into the test failure. Looks like it's failing in Firefox due to some interaction between the shady dom polyfill and the focus-visible polyfill.

cdata:
I have addressed the test failures. The following changes were necessary:

 - In some browsers, we accessed the `content` of the inline `<template>` in the Shadow DOM test fixture too early. Deferring until `DOMContentLoaded` fixed the issue.
 - I had (mistakenly) assumed that the polyfill did not need to be applied at `shadowRoot` boundaries when the ShadyCSS polyfill was in effect. But, it turns out the polyfill does encapsulate focus events at the virtual shady root boundaries that it creates, so the `:focus-visible` polyfill should still be applied as normal.

src/focus-visible.js
cdata: I switched the order of the conditions around. The comments are hopefully more cleanly formatted now.

alice:
Aww, no commit history? We squash merge anyway.

cdata:
Sorry, I will try to recreate the history

alice:
Thanks and sorry :\

cdata:
Nothing a little git gymnastics couldn't handle 👍

I split the change out into two commits. It looks like the Node.js 10 build had three test timeouts this time around. I'm not sure if that is actionable...

alice:
Thanks so much for the git gymnastics! Build problems look like they were an infrastructure problem, it passed once I restarted it.

cdata:
Thanks @alice ! BTW I should note before anyone merges this that due to the fix for #196 this should probably be considered a breaking change (albeit a small one). The timing around which the polyfill activates is different now, so it could break any user who is sensitive to that.

test/fixtures/shadow-dom.html
```diff
@@ -36,18 +36,21 @@
...
+        el.attachShadow({ mode: 'open', delegatesFocus: true });
+        el.shadowRoot.appendChild(
+            shadowRootTemplate.content.cloneNode(true));
 
-      // If ShadyCSS is being used, there is no need to apply the polyfill to
-      // the shadow root because the shadow tree is not truly encapsulated:
-      if (self.ShadyCSS == null || self.ShadyCSS.nativeShadow) {
         applyFocusVisiblePolyfill(el.shadowRoot);
-      }
+      }, { once: true });
```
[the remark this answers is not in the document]
cdata: 🎉

cdata:
@robdodson LMK if I can do anything to help move this one forward!

image

README.md
```diff
@@ -95,6 +95,42 @@ You can use a service like [Polyfill.io](https://polyfill.io) to download only t
...
+### Shadow DOM
+
+It could be very expensive to apply this polyfill automatically to every shadow
+root that is created in a given document, so the polyfill ignores shadow roots
+by default. If you are using Shadow DOM in a component, it is possible to apply
+this polyfill imperatively to the component's shadow root:
+
+```javascript
+// Check for the polyfill:
+if (self.applyFocusVisiblePolyfill != null) {
```
robdodson: Is `self` equivalent to `window`? Not sure if the snippet makes it clear what context this is running in.

README.md
```diff
@@ -95,6 +95,42 @@ You can use a service like [Polyfill.io](https://polyfill.io) to download only t
...
+When this polyfill is lazy-loaded, and you are applying the polyfill to a shadow
+root with JavaScript, it is important to know when the polyfill has become
+available before trying to use it.
+
+In order to act at the right time, you can observe the global
+`focus-visible-polyfill-ready` event:
+
+```javascript
+window.addEventListener('focus-visible-polyfill-ready',
+    () => self.applyFocusVisiblePolyfill(myComponent.shadowRoot),
```
robdodson: Similar to above, what's the expected value of `self` in this context?

README.md
cdata: Yeah, `self` is equivalent to `window` in the document context. I can change this to `window`.

README.md
cdata: Will change to `window`.

robdodson:
I think this is good to merge once chris makes the `window` changes in the README.

cdata:
@robdodson README has been updated per your request 🍻

src/focus-visible.js
```diff
@@ -230,45 +234,58 @@ function init() {
...
-    callback();
-  } else {
-    loaded = false;
-    document.addEventListener('DOMContentLoaded', load, false);
-    window.addEventListener('load', load, false);
-  }
+try {
+  event = new CustomEvent('focus-visible-polyfill-ready');
+} catch (error) {
+  // IE11 does not support using CustomEvent as a constructor directly:
```
jakearchibald: Does it support `new Event()`?

cdata: Some very light Googling yields this [Stack Overflow](https://stackoverflow.com/questions/26596123/internet-explorer-9-10-11-event-constructor-doesnt-work) that suggests "no" and "use CustomEvent instead"  😄

PR #197, opened 2019-05-20
IE11: ensure target.nodeName exists

thebuilder:
I've seen a few cases where IE11 throws this error:
```
Unable to get property 'toLowerCase' of undefined or null reference
```

See full example on Sentry: https://sentry.io/share/issue/f02ceb35bf504436a21a60c08154341d/

The `e.target.nodeName` in `onInitialPointerMove` is `undefined`, so attempting to read it throws an error.

This pull request adds an extra guard that checks if `nodeName` exists, before calling `toLowerCase()`.


robdodson:
LGTM. Restarting the travis build to see if i can make it happy.

PR #200, opened 2019-06-18
Add guard for SSR environments

### LINKED ISSUE #199: v5 breaks server-side rendering
We use `focus-visible` in a server-side rendered Next.js application.

Prior to v5, we could simply `import 'focus-visible';` at the top of our `_app.js`, which is the entry-point of all pages, and where we can setup the Layout and other things.

Since upgrading from `v4.1.5` to `v5.0.0` the development server breaks with the following message:
```sh
/webapp/node_modules/focus-visible/dist/focus-visible.js:1
ReferenceError: window is not defined
    at /webapp/node_modules/focus-visible/dist/focus-visible.js:285:3
    at window.applyFocusVisiblePolyfill (/webapp/node_modules/focus-visible/dist/focus-visible.js:2:66)
    at Object.<anonymous> (/webapp/node_modules/focus-visible/dist/focus-visible.js:5:2)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
```

I believe this might have been caused by [the following changes](https://github.com/WICG/focus-visible/compare/v4.1.5...v5.0.0#diff-bf74a65b572b79d701701a8b9b1e01aeL252):
image

cdata:
This wraps the final interactions with `window` and `document` respectively to guard against the case that they are not there (ostensibly this is the case in SSR scenarios).

Fixes #199

### DIFF
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -273,25 +273,32 @@ function applyFocusVisiblePolyfill(scope) {
   }
 }
 
-// Make the polyfill helper globally available. This can be used as a signal to
-// interested libraries that wish to coordinate with the polyfill for e.g.,
-// applying the polyfill to a shadow root:
-window.applyFocusVisiblePolyfill = applyFocusVisiblePolyfill;
+// It is important to wrap all references to global window and document in
+// these checks to support server-side rendering use cases
+// @see https://github.com/WICG/focus-visible/issues/199
+if (typeof window !== 'undefined') {
+  // Make the polyfill helper globally available. This can be used as a signal
+  // to interested libraries that wish to coordinate with the polyfill for e.g.,
+  // applying the polyfill to a shadow root:
+  window.applyFocusVisiblePolyfill = applyFocusVisiblePolyfill;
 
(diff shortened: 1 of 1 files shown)


robdodson:
LGTM. Before merging, @damusnet could you try this in your project to confirm if fixes your issue?

damusnet:
@robdodson would this be published on npm anywhere as a pre-release or something?

I can try and update my local code to reflect the changes otherwise.

damusnet:
Alright, so this is far from scientific, but I copy/pasted the 2 `if (typeof` changes into my `node_modules/focus-visible/dist/focus-visible.js` after upgrading to v5, and long story short it works for me! Thanks for being so responsive!

(and I'm super excited by this v5 because I love that alt/ctrl/cmd key do not trigger the outline anymore)

cdata:
Thanks for testing that out @damusnet

robdodson:
coool thanks folks. Just rolling out v5.0.1 now :)

PR #202, opened 2019-07-01
Add document guard for SSR

josemarluedke:


robdodson:
@josemarluedke are there SSR scenarios where `window` is defined by `document` isn't?

josemarluedke:
@robdodson I found that strange as well, but it was the case when using Ember FastBoot.

robdodson:
ah ok, thanks for that!

PR #206, opened 2019-08-13
Remove redundant call to clearTimeout

oliviertassinari:
I was working on https://github.com/mui-org/material-ui/issues/16976, by luck, the concern was already taking care of in https://github.com/WICG/focus-visible/issues/167. In the process of applying the diff, I have found this strange call to clearTimeout—I believe it's not needed.


alice:
Yeah, it does seem like we could only possibly need one of those...

Issue #179, opened 2018-08-31
Conflict with how Vue handles class bindings, and potential workaround


tinymachine:
In Vue 2.5.17, when I [bind classes](https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax) to an element that can receive focus, Vue appears to overwrite the `.focus-visible` class added by this polyfill. Thankfully the `data-focus-visible-added` attribute is also added by the polyfill, and that attribute is left untouched, so a workaround is simply to change the [CSS rule recommended in the README](https://github.com/WICG/focus-visible/tree/v4.1.5#2-update-your-css) (as of 4.1.5) from:

```css
.js-focus-visible :focus:not(.focus-visible) {
  outline: none;
}
```

to:

```css
.js-focus-visible :focus:not([data-focus-visible-added]) {
  outline: none;
}
```

Is it safe to assume that this data attribute will continue to be added in updates to the polyfill?
(Thanks for creating and sharing this polyfill, btw!)

robdodson:
I think we can add that as an additional option to the README. The attribute is really an implementation detail, we use it to detect if the class was added by the polyfill (instead of manually by the author).

As much as possible we'd like folks to just use the class but I totally understand why that won't work in this case. I'll add a note to the README.

robdodson:
Added a PR over in #180. Thanks for reporting it!

tinymachine:
That's great -- hope that can be of help to others. BTW I'm a big fan of your accessibility work, and as someone new to really digging into web accessibility, I've found your YouTube series [A11ycasts](https://www.youtube.com/playlist?list=PLNYkxOF6rcICWx0C9LVWWVqvHlYJyqw7g) extremely helpful and engaging. Kudos, and keep up the excellent work!

robdodson:
Thank you! Glad you're finding the videos helpful 😁

Malvoz:
With concerns around the removal of classes, a more robust approach would be to leave out class selectors entirely:

```CSS
[data-js-focus-visible] :focus:not([data-focus-visible-added]) {
  outline: none;
}
```

PR #219, opened 2020-03-08
Set an attribute on the document in addition to the class

ivoilic:
Currently only a class is added to the document. This makes it difficult if you are using a  framework which overwrites your classes. This mirrors the solution that is already in place for this same issue. See issue [#179](https://github.com/WICG/focus-visible/issues/179).

### DIFF
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -269,6 +269,7 @@ function applyFocusVisiblePolyfill(scope) {
     scope.host.setAttribute('data-js-focus-visible', '');
   } else if (scope.nodeType === Node.DOCUMENT_NODE) {
     document.documentElement.classList.add('js-focus-visible');
+    document.documentElement.setAttribute('data-js-focus-visible', '')
   }
 }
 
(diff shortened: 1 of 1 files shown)


yoavweiss:
@ivoilic Can you please join the WICG to appease the IPR bots?

ivoilic:
@yoavweiss I've joined WICG and updated the PR due to a typo. Sorry for not doing so before I submitted the PR.

yoavweiss:
@ivoilic no worries :) The IPR bots are still not happy. Did you link your GH account with your W3C account?

ivoilic:
@yoavweiss Yes, I just forced it to check again and it looks like it went through!

ivoilic:
Any chance of getting this merged? This is a change that shouldn't have a negative impact on existing user and would be of great benefit to many.

robdodson:
published a v5.1.0 release with this in it. thank you!

ivoilic:
@robdodson 🎉 Great! Although I just realized I probably should have updated the docs in this PR as well.

PR #226, opened 2020-05-03
Reduce bundle size by defining pointerMoveEvents array

wojtekmaj:
Instead of 18 almost identical lines, use pointerMoveEvents array to keep track of all listeners to add/remove in addInitialPointerMoveListeners and removeInitialPointerMoveListeners respectively.

### DIFF
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -191,6 +191,18 @@ function applyFocusVisiblePolyfill(scope) {
     }
   }
 
+  const pointerMoveEvents = [
+    'mousemove',
+    'mousedown',
+    'mouseup',
+    'pointermove',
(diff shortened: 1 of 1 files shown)


Justineo:
Have you compared the gzipped size of the bundle? “Almost identical” means you’ll have a very good compression ratio.

yoavweiss:
@wojtekmaj - can you join the WICG to appease the IPR bots? Thanks! :)

robdodson:
hm it seems like a bunch of tests are failing. Were they passing for you locally? I'm wondering if some part of our setup is busted.

wojtekmaj:
Looks like the tests are failing because Array.forEach is not supported on IE.

PR #229, opened 2020-05-16
Enable button tagName in focusTriggersKeyboardModality

skosito:


skosito:
I see this broke some tests... If you guys think that this can be ok for patch, I can fix tests. If not, do you mind if I patch this change using patch-package and use it in codebase like that? Are there some side effects to this change? Personally I do not see a reason why button focus should not trigger switching to keyboard modality, but amount of failed tests worries me a bit...

eps1lon:
On many platforms clicking a button also focuses this button. You wouldn't want to enter keyboard modality in these situations. This is what the test are verifying: https://travis-ci.org/github/WICG/focus-visible/jobs/687664906#L351

skosito:
@eps1lon Thanks for reply. But what about situation where I am going from url bar to button as a first element? Sometimes button is skipped because of this.

yoavweiss:
@skosito - Can you join the WICG to appease the IPR bots?

robdodson:
Thanks for the PR @skosito 

It's intentional that we don't match focus-visible when you mouse click on a button. We mention this [in the explainer](https://github.com/WICG/focus-visible/blob/master/explainer.md#rationale).

If you'd like to always apply a focus style to buttons, regardless of if the user used a mouse or keyboard, you can use the regular `:focus` pseudo class.

skosito:
Thank you very much for help!

PR #245, opened 2020-11-05
fix: Tolerate multiple imports/loads of the script

maranomynet:
It may come up that multiple scripts import or load the polyfill into the same document.

In such cases we want to ensure the polyfill only runs once.

### DIFF
--- a/src/focus-visible.js
+++ b/src/focus-visible.js
@@ -276,7 +276,7 @@ function applyFocusVisiblePolyfill(scope) {
 // It is important to wrap all references to global window and document in
 // these checks to support server-side rendering use cases
 // @see https://github.com/WICG/focus-visible/issues/199
-if (typeof window !== 'undefined' && typeof document !== 'undefined') {
+if (typeof window !== 'undefined' && typeof document !== 'undefined' && !window.applyFocusVisiblePolyfill) {
   // Make the polyfill helper globally available. This can be used as a signal
   // to interested libraries that wish to coordinate with the polyfill for e.g.,
   // applying the polyfill to a shadow root:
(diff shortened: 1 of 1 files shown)


src/focus-visible.js
```diff
@@ -295,9 +295,7 @@ if (typeof window !== 'undefined' && typeof document !== 'undefined') {
   }
 
   window.dispatchEvent(event);
-}
 
-if (typeof document !== 'undefined') {
```
maranomynet, line 300: I'm assuming these conditional blocks can be safely joined, for simplicity+brevity

src/focus-visible.js
```diff
@@ -276,7 +276,7 @@ function applyFocusVisiblePolyfill(scope) {
 // It is important to wrap all references to global window and document in
 // these checks to support server-side rendering use cases
 // @see https://github.com/WICG/focus-visible/issues/199
-if (typeof window !== 'undefined' && typeof document !== 'undefined') {
+if (typeof window !== 'undefined' && typeof document !== 'undefined' && !window.applyFocusVisiblePolyfill) {
```
maranomynet, line 279: Just checking for `applyFocusVisiblePolyfill` is sufficiently safe as the function is always run right after it's assignment into the global Window scope

maranomynet, line 279: Just checking for `applyFocusVisiblePolyfill` is sufficiently safe as the function is always run right after it's assignment into the global Window scope

yoavweiss:
@maranomynet - can you join the WICG to appease the IPR bots?

maranomynet:
I've ventured down that rabbit hole of bad onboarding UX, and emerged mostly unscathed.

* [x] IPR bots appeased

maranomynet:
I can't believe it's been three years already. Time flies.

PR #253, opened 2021-04-03
Update README.md

### LINKED ISSUE #252: Add hint to documentation about ES6 import
I think it would be practical to have a hint about how to use the script with ES6 imports.

It would be great to have a hint in the docs, that says:

**You do not have to use a script tag but can directly import the script as follows:

import "focus-visible"**

benjaminpreiss:
Add hint for ES6 imports to Readme

fixes #252


ElenaZimina:
For me it works.

